PR #2, opened 2019-11-21
Marshmallow

FynnBe:
use to Marshmallow schema validation


oeway:
Hey, would it be possible to export a json schema (https://json-schema.org/implementations.html) file so we can use it in JS for the web dev tool and also in JAVA?

I did a quick check and it seems pretty straightforward with Marshmallow: https://github.com/fuhrysteve/marshmallow-jsonschema

pybio/spec/fields.py
```diff
@@ -22,9 +23,58 @@ def resolve_local_path(path_str: str, context: dict) -> pathlib.Path:
...
+    if uri.scheme.lower() != "doi" and not uri.netloc.endswith("doi.org"):
+        return uri
+
+    doi = uri.path.strip("/")
+
+    url = "https://doi.org/api/handles/" + doi
+    r = json.loads(requests.get(url).text)
+    response_code = r["responseCode"]
+    if response_code != 1:
+        raise RuntimeError(f"Could not resolve doi {doi} (responseCode={response_code})")
```
MartinHjelmare: It's nice for the consumer of this library if we create library specific exceptions and raise as needed. Create a library specific exception that inherits `Exception` and make further child exceptions from that parent as needed.

pybio/spec/fields.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+def resolve_doi(uri: ParseResult) -> ParseResult:
+    if uri.scheme.lower() != "doi" and not uri.netloc.endswith("doi.org"):
+        return uri
+
+    doi = uri.path.strip("/")
+
+    url = "https://doi.org/api/handles/" + doi
+    r = json.loads(requests.get(url).text)
```
MartinHjelmare:
```suggestion
    r = requests.get(url).json()
```

pybio/spec/fields.py
```diff
@@ -0,0 +1,184 @@
...
+from urllib.parse import urlparse, ParseResult
+import json
+import pathlib
+import requests
+import subprocess
+import typing
+import yaml
+
+from marshmallow import ValidationError
+from marshmallow.fields import *
```
MartinHjelmare: Can we be explicit and import the names we need? It's easier to read then.

FynnBe: changed. (I wanted fields to be an extension of marshmallow.fields such that in schema.py fields.AnyField can be used).

pybio/spec/fields.py
FynnBe: good idea, I started this in https://github.com/bioimage-io/python-bioimage-io/pull/2/commits/0d0626552f699b77c858ec6d9e9f62ea3304857c

pybio/spec/pybio_types.py
```diff
@@ -0,0 +1,167 @@
+from dataclasses import dataclass
```
MartinHjelmare: We'll require Python 3.7 when using dataclasses, right?

pybio/spec/schema.py
```diff
@@ -0,0 +1,250 @@
...
+    doi = fields.Str(missing=None)
+    url = fields.Str(missing=None)
+
+    @validates_schema
+    def doi_or_url(self, data, **kwargs):
+        if data["doi"] is None and data["url"] is None:
+            raise ValidationError("doi or url needs to be specified in a citation")
+
+
+class MinimalYAML(Schema):
```
MartinHjelmare: Will it be confusing at some point that there is a schema and dataclass named the same? Maybe we're never using them together?

FynnBe:
> Hey, would it be possible to export a json schema (https://json-schema.org/implementations.html) file so we can use it in JS for the web dev tool and also in JAVA?
> 
> I did a quick check and it seems pretty straightforward with Marshmallow: https://github.com/fuhrysteve/marshmallow-jsonschema

I guess that would be possible with some custom type mapping through https://github.com/fuhrysteve/marshmallow-jsonschema#custom-type-support

In any case, let's first polish the pure marshmallow version. 
I'm not so sure about the Magic Enum values, etc...

pybio/spec/schema.py
```diff
@@ -0,0 +1,250 @@
...
+            if not (k in spec.required_kwargs or k in spec.optional_kwargs):
+                raise ValidationError
+
+
+class InputShape(Schema):
+    # exact = fields.List(fields.Integer(), missing=None)
+    min = fields.List(fields.Integer(), required=True)
+    step = fields.List(fields.Integer(), required=True)
+
+    # @validates_schema
```
MartinHjelmare: Commented code.

pybio/spec/source.py
```diff
@@ -0,0 +1,273 @@
...
+        assert self.cache_path in self.local_file_path.parents, "only download files to cache folder!"
+
+        uri = self.uri
+        if self.download_commands:
+            for command in self.download_commands:
+                call(command)
+
+            assert self.local_file_path.exists(), self.local_file_path
+            if self.object_name and self.object_name[0] == "L" and self.object_name[1:].isdigit():
+                # get object name from line number
```
MartinHjelmare: Why do we care about line number? Is that part of the spec?

MartinHjelmare: Is this to allow reading parts of modules hosted on github?

MartinHjelmare:
Adding a test where the library is used would be nice, also as an example for readers of the code. I recommend pytest as test runner.

pybio/spec/pybio_types.py
FynnBe: yes, is that a problem?

pybio/spec/schema.py
FynnBe: at the moment they are all mirrored (there is also code for autogenerating schemas from dataclasses: https://pypi.org/project/marshmallow-dataclass/, but not all fields would map one to one)

pybio/spec/source.py
FynnBe: yes, the fragment of a github url may contain the line number of the definition of that object. I thought this would be a neat way of defining importable objects... Having said that, this is deprecated, as we only get at yaml file here (and importable objects should always be specified as dependencies).

FynnBe: ah, this is in source.py... this whole file is deprecated... I'll delete the old files..

pybio/spec/pybio_types.py
MartinHjelmare: Not specifically I think. It depends on what environments we want to support and how easy to make it for users.

- Ubuntu 18.04 still defaults to Python 3.6. Those users will  need to install Python 3.7 eg from a ppa.
- Debian Buster comes with Python 3.7.
- OS X doesn't come with Python 3 at all, as far as I know, so it doesn't matter here.
- Windows users will also need to install Python manually.
- Google colab only supports Python 3.6 so far.

Do we see any limitations with the above?

MartinHjelmare: - Ubuntu 20.04 will have Python 3.8.

constantinpape: @MartinHjelmare thanks for bringing this up, these are valid concerns.
We are a bit biased here by having all our workflows on conda, which makes switching to newer python versions relatively easy.
However, loosing compatibility with the default ubuntu python version and the default for colab is certainly something to consider, because there might be a potential user base who uses these set ups with system python.
So I think it depends on whether we want to limit support to conda. If this is the case, we can safely go with python 3.7. If we want to support system python + pip though, support for python 3.6 would be desirable.

MartinHjelmare: I like using newer Python versions too.

@oeway does it matter from your perspective?

oeway: I think it would be great if we make it compatible with python 3.6, considering this going to be a shared library. For BioEngine for example, we would like to run it in google colab.

MartinHjelmare: There seems to be a backport of dataclasses specifically for Python 3.6. We'd have to install it only on Python 3.6 I guess.
https://github.com/ericvsmith/dataclasses

pybio/spec/schema.py
```diff
@@ -0,0 +1,253 @@
+from dataclasses import asdict
+from typing import Any, Dict, Union
+
+from marshmallow import Schema as _Schema, pprint, ValidationError, post_load, validates_schema, validates
+
+from pybio.spec import pybio_types, fields
+from pybio.spec.pybio_types import MagicTensorsValue, MagicShapeValue
+
+
+class Schema(_Schema):
```
constantinpape: Is it maybe a bit confusing that we call this `Schema` and inherit from `marshmallow.Schema`?
We could call it `PybioSchema` and then also avoid the `from marshmallow import Schema as _Schema`.

pybio/spec/schema.py
```diff
@@ -0,0 +1,253 @@
...
+    @post_load
```
constantinpape: If I understand this correctly, the `post_load` decorator will automatically return the corresponding dataclass
when deserializing a `Schema`. Do we want that? Having access to the `Schema` would allow to use all the fancy `dump()` functionality that marshmallow provides directly.
We could just leave this as a normal method that returns the dataclass.

pybio/spec/fields.py
```diff
@@ -0,0 +1,186 @@
...
+            if uri.netloc:
+                raise PyBioValidationException(f"Invalid URI: {uri}")
+            spec_path = resolve_local_path(uri.path, self.context)
+        elif uri.netloc == "github.com":
+            orga, repo_name, blob, commit_id, *in_repo_path = uri.path.strip("/").split("/")
+            in_repo_path = "/".join(in_repo_path)
+            cached_repo_path = self.cache_path / orga / repo_name / commit_id
+            spec_path = cached_repo_path / in_repo_path
+            if not spec_path.exists():
+                cached_repo_path = cached_repo_path.resolve().as_posix()
```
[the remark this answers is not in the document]
FynnBe: yes
```python
from pathlib import Path
a = Path("/lala/here and there/lülö-da")
a
Out[4]: WindowsPath('/lala/here and there/lülö-da')
a.as_posix()
Out[5]: '/lala/here and there/lülö-da'
Path(a.as_posix())
Out[6]: WindowsPath('/lala/here and there/lülö-da')
```

pybio/exceptions.py
```diff
@@ -0,0 +1,13 @@
+from marshmallow import ValidationError
+
+
+class PyBioException(Exception):
```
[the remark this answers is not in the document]
constantinpape: I am resolving this for now, see my general comments for more detail.

pybio/spec/fields.py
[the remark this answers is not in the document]
constantinpape: Sorry, my initial comment was not quite clear. 
I was worrying if this works in the `subprocess.call` to `git`, because it would be something like this:
```
git clone reponame.git /path/to/repo
```
Where `/path/to/repo` is encoded as posix path, which might cause problems on Windows?

constantinpape:
@FynnBe @MartinHjelmare @oeway 

Just wanted to get your opinion on how to move along with this.
From what I gather from the comments, everyone is in principle ok with this approach using `marshmallow`.
There's of course a lot of things left to do, including points we all brought up:
- Adding tests and examples
- Making the code easier to understand by adding some comments and doc-strings
- Exporting a json schema

But I think if we all agree on the overall approach, it might be better to merge this PR first and then address these things in subsequent PRs, which would make reviewing easier.
Let me know what you think.

MartinHjelmare:
Just writing down some thoughts I had about marshmallow.

I'm not familiar with marshmallow and haven't looked into it so much yet. We need validation and serialization. Was there other requirements in the plan when settling on marshmallow?

Using classes for output schemas, as marshmallow does, makes the code pretty verbose but it also gives some features that may be good to have.

The documentation looks solid. I'm ok with using marshmallow. :+1:

oeway:
Sounds great! I haven't had time to go through the code in greater details, however, I think marshmallow seems good, and we should move forward to merge the PR if it's already being reviewed.

PR #3, opened 2019-12-05
ImportableField should use spec path for module discovery

m-novikov:
When I pack model.py in the same directory as yaml spec I expect to be able to import it.


FynnBe:
does this PR collide with #8 ?
otherwise LGTM

PR #7, opened 2020-01-13
Passing tests

FynnBe:
transformation improvements and clean up
related to https://github.com/bioimage-io/pytorch-bioimage-io/pull/7


oeway:
Great Job, @FynnBe!

Regarding transformations and losses, I think an important difference is we do not track the gradient of transformations used for pre- and post- processing, but we do need that for losses. In that sense, maybe we should separate the two?

Also, a random idea, perhaps we should provide a generic transformation that automatically maps to Tensorflow/pytorch ops/layers?

FynnBe:
> Overall this makes sense. I am not quite sure how well the transformation and loss logic maps
> to other frameworks than pytorch, so I am not really sure if we should have any implementations at this level.
> 
> This depends on how much flexibility we want other implementations to have, probably it's best to stick to the solution we have right now and reconsider this if there are some major issues preventing adoption of this.

Right, we talked about this before the break, sorry I forgot some details, I meant to take out the loss part from pybio...

FynnBe:
> Great Job, @FynnBe!

thanks

> Regarding transformations and losses, I think an important difference is we do not track the gradient of transformations used for pre- and post- processing, but we do need that for losses. In that sense, maybe we should separate the two?

exactly! Constantin concluded this already before, but I forgot over the winter break. 

> Also, a random idea, perhaps we should provide a generic transformation that automatically maps to Tensorflow/pytorch ops/layers?

if I understand you correctly, transformations in pybio will do just that -- based on framework independent implementations (numpy, etc) -- whereas 'loss transformations' need to be implemented per framework. Sharing any code across frameworks (have a loss base in pybio, as I added it in atm (will move it to torchbio) only entangles the frameworks unnaturally...

pybio/core/transformations/reshape.py
```diff
@@ -9,7 +11,7 @@ def __init__(self, shape: Sequence[int], **super_kwargs):
         self.shape = shape
```
constantinpape: Do we have two `Reshape` transformations now? One from `numpylike` and the one here?

FynnBe: yes, however there is only one spec (for the non-numpylike one)

FynnBe: I guess we could just remove the numpylike Reshape, I left it in as its a reasonable numpylike one... (and allows for more [specific] kwargs)

PR #9, opened 2020-02-01
Consistent variable names and SpecURI

FynnBe:
To enable the `URITransformer` to descend into referenced specs I added `SpecURI`. 
While doing so I appreciated @m-novikov 's addition of the `Node` class and noticed some older confusing variable names. 
Here I propose the following name changes (all changes in backend only):
Fundamental name changes:
- module `pybio.spec.spec_types` -> `pybio.spec.node` (it specifies all Nodes)
- **BaseSpec**(Node) -> **SpecWithKwargs**(Node)
- **ModelSpec**(BaseSpec) -> **Model**(SpecWithKwargs) (also for Transformation, Reader, Sampler)
- **Model**(MinimalYAML, WithInputs, WithOutputs) -> **ModelSpec**(BaseSpec, WithInputs, WithOutputs) (also for Transformation, Reader, Sampler)
-  MinimalYAML(Node) -> BaseSpec(Node, WithSource)

Especially the **highlighted** changes are motivated by the following train of thought:
A spec like <name>.transformation.yaml now is represented by the class `TransformationSpec(BaseSpec)`. If used somewhere else (e.g. in a <name>.model.yaml) a `transformation` is a pair of `spec` and `kwargs`, thus represented by `Transformation(SpecWithKwargs)`.
Model specs are somewhat special, as their checkpoints are generally only valid for the default values of optional kwargs, as a `ModelSpec` instance is equivalent to a `Model'='ModelSpecWithKwargs' without any given kwargs. Generally overwriting any optional kwargs might break something. I still think we should keep this pattern (at least for the forseeable future). If we want to emphasize the danger of loading models with 'new' kwargs, we might consider to remove `kwargs` from the `load_model(spec_uri: str, **kwargs)` function signature.

Derived changes:
- `model_spec` -> `pybio_model`

side note: Furthermore I changed the `Importable`s to inherit from `Importable` instead of being member classes to allow for correct type checking (fixes type hint in https://github.com/bioimage-io/python-bioimage-io/blob/c355c4e82dc6b2c0f3394853d04f72e5ffb1bbdc/pybio/spec/utils.py#L48):
- `spec_types.Importable.Module` -> `node.ImportableFromModule(Importable)`
- `spec_types.Importable.Path` -> `node.ImportableFromPath(Importable)`

### DIFF
--- a/pybio/core/training/classic_fit.py
+++ b/pybio/core/training/classic_fit.py
@@ -1,14 +1,14 @@
-from pybio.spec.spec_types import ModelSpec
+from pybio.spec.node import Model
 from pybio.spec.utils import get_instance
 
 
-def classic_fit(model_spec: ModelSpec, start: int = 0, batch_size: int = 1):
+def classic_fit(pybio_model: Model, start: int = 0, batch_size: int = 1):
     """classic fit a la 'model.fit(X, y)'"""
 
-    model = get_instance(model_spec)
-    reader = get_instance(model_spec.spec.training.setup.reader)
-    sampler = get_instance(model_spec.spec.training.setup.sampler, reader=reader)
+    model = get_instance(pybio_model)
+    reader = get_instance(pybio_model.spec.training.setup.reader)
+    sampler = get_instance(pybio_model.spec.training.setup.sampler, reader=reader)
     X, y = sampler[start, batch_size]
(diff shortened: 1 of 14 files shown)


m-novikov:
Resolved by #10

PR #17, opened 2020-02-12
add SourceTransformer

FynnBe:


pybio/spec/utils.py
```diff
@@ -67,17 +73,12 @@ def transform_list(self, node: list) -> list:
...
-        module = importlib.import_module(importable.module_name)
-        return getattr(module, importable.callable_name)
-    elif isinstance(importable, nodes.ImportablePath):
-        importlib_spec = importlib.util.spec_from_file_location(f"user_imports.{uuid.uuid4().hex}", importable.filepath)
-        dep = importlib.util.module_from_spec(importlib_spec)
-        importlib_spec.loader.exec_module(dep)
-        return getattr(dep, importable.callable_name)
+@dataclasses.dataclass
+class ImportedSource:
+    callable_: Callable
```
m-novikov: Maybe factory is a better name? It communicates that this callable used to create objects.

pybio/spec/utils.py
```diff
@@ -96,6 +97,11 @@ def get_instance(node: Union[nodes.SpecWithKwargs, nodes.WithImportableSource],
 
         return get_instance(node.spec, **joined_spec_kwargs)
     elif isinstance(node, nodes.WithImportableSource):
+        if not isinstance(node.source, ImportedSource):
+            raise PyBioTransformationOrderException(
```
m-novikov: I think there is no need for the custom exception type here. We only need custom exceptions when user can catch them explicitly to take some action, but in this case it looks like ValueError with same error message is a better fit.

pybio/spec/utils.py
FynnBe: agreed

pybio/spec/utils.py
FynnBe: Alright, we might want to emphasize in the module api that load_spec_with_kwargs and get_instance are the appropriate funcitons to use... they are a bit hidden away in spec.utils atm

Issue #31, opened 2020-06-25
Key word arguments for model specs


FynnBe:
Should we add optional and required key word arguments to the model specification, as we have them for other specifications, e.g. reader, sampler, transformation?

(assumption: key word arguments have to be json serializable, no fancy stuff!)

Pros:
- consistency across specs
- most models actually do have some kwargs, like nr of channels, etc... 

Cons:
- we would have to specify model kwargs for the different weights to avoid collisions

constantinpape:
This is outdated.

Issue #27, opened 2020-03-05
Export JSON schema for model specification


FynnBe:

oeway:
Is this already done?


I just check the schema, and noticed there are wired keys appeared in the root: `optional_kwargs` `required_kwargs`, why is that? @FynnBe 

I am thinking maybe we should write the schema manually, in that case, will you be able to read a json schema file and use it in the python package?

FynnBe:
> Is this already done?

yes, I forgot about this issue... 

> I just check the schema, and noticed there are wired keys appeared in the root: optional_kwargs required_kwargs, why is that? @FynnBe

The `optional_kwargs` and `required_kwargs` keys are there to specify any kwargs optional/required for model instantiation, as source points to a callable... 

> I am thinking maybe we should write the schema manually, in that case, will you be able to read a json schema file and use it in the python package?

we use marshmallow to specify the schema, which allows for more validation and exporting of the json schema (with a subset of the 'rules'). As marshmallow is more powerful, I would like to keep it this way. 

What would you want to change manually? Maybe we can just adapt the schema in marshmallow accordingly?

oeway:
Thanks @FynnBe  for the explanation.

> The `optional_kwargs` and `required_kwargs` keys are there to specify any kwargs optional/required for model instantiation, as source points to a callable...

But this is not in our configuration specification right, is there a way to clean up the exported json schema?

> What would you want to change manually? Maybe we can just adapt the schema in marshmallow accordingly?

I would like to make a model config generator/validator plugin in javascript, so users can build model config from an UI and we can add a badge for validated model cards. Even more, for those not validated ones we show an info button where the developer can click and we can show a list of errors, warnings and suggestions.

I can either redo it in javascript and make an bioengine app for that, or we can run the marshmallow validation when we compile the yaml file to the manifest json (in a CI), then add the validation report to the generated json file.

The JSON schema is also important, even though it's a subset of the rules, we can share it between different language (e.g. Java/JS).

Related to these, I have two additional questions:
1. when I pass the yaml config to marshmallow to validate, can you generate a detailed report? Not just passed or failed, but ideally also more detailed information for different levels: error, warning, recommendation etc.

2. For the generated JSON schema, can we clean it up to remove for example keys like `required_kwargs`, it's a bit confusing when we generate an HTML form with it and it has these keys.

FynnBe:
> But this is not in our configuration specification right, is there a way to clean up the exported json schema?

[...]

PR #32, opened 2020-06-25
drop model kwargs in json schema

FynnBe:
model key word arguments were not part of the spec and only implemented in the Marshmallow schema. 
This PR removes them from the generated JSON schema, as discussed in #27 
If we want to add model key word arguments to the model specification can be discussed in #31


oeway:
Looks good.

PR #34, opened 2020-10-20
update to format 0.3.0

FynnBe:


pybio/core/datasets/base.py
```diff
@@ -1,16 +1,18 @@
-from typing import Sequence, Tuple
+from typing import Optional, Sequence, Tuple
+
+import torch.utils.data
```
m-novikov, line 3: We should avoid framework-specific import in the core library.

PR #35, opened 2020-10-29
Verify spec

FynnBe:
also contains:
- changes to adhere to format version 0.3.0, which is not yet finalized.
- draft of a runner for scikit-learn models (we are not planning to officially support sklearn models at this point)


oeway:
Nice, we should also add this verification to the compilation script and generate a report for each model (e.g. highlight as an error icon on the model card if error detected).

oeway:
@FynnBe Can you generate an error report after verification, not only throw error but output details about the error?

FynnBe:
> @FynnBe Can you generate an error report after verification, not only throw error but output details about the error?

example from travis:
image

constantinpape:
The checker works for my most current v3 spec:
https://github.com/hci-unihd/batchlib/blob/master/misc/models/torch/bioimageio/UNetCovidIf.model.yaml

What's left to do before merging?

FynnBe:
> What's left to do before merging?

get the tests to pass...

pybio/core/datasets/base.py
```diff
@@ -0,0 +1,34 @@
+import collections
+from typing import Dict, Optional, OrderedDict, Sequence, Tuple
```
m-novikov, line 2: Use Dict instead of OrderedDict

FynnBe: I went with MutableMapping instead, but seem to have forgotten it here... I'll fix it

m-novikov: I don't know if MutableMapping implies ordering, while python Dict does.

FynnBe:
I am a bit puzzled why this fails:
```
self = PosixPath('/home/travis/build/bioimage-io/python-bioimage-io/tests/conftest.py/../../specs/models/sklearn/RandomForestClassifier_v0_1.model.yaml')
name = '/home/travis/build/bioimage-io/python-bioimage-io/tests/conftest.py/../../specs/models/sklearn/RandomForestClassifier_v0_1.model.yaml'
flags = 524288, mode = 438
    def _opener(self, name, flags, mode=0o666):
        # A stub for the opener argument to built-in open()
>       return self._accessor.open(self, flags, mode)
E       NotADirectoryError: [Errno 20] Not a directory: '/home/travis/build/bioimage-io/python-bioimage-io/tests/conftest.py/../../specs/models/sklearn/RandomForestClassifier_v0_1.model.yaml'
```

FynnBe:
> I am a bit puzzled why this fails:

for PosixPaths this does not seem legal: `posix_path/a_file.ext/../elsewhere`

PR #53, opened 2021-03-12
add run_mode to spec

FynnBe:


m-novikov:
Looks great, but would be nice to have tests for this [see here](https://github.com/bioimage-io/python-bioimage-io/blob/master/tests/test_pybio/test_spec/test_schema.py#L133)

PR #61, opened 2021-04-26
generating spec doc from spec validation code

FynnBe:
Motivated by https://github.com/bioimage-io/spec-bioimage-io/issues/74 here is a draft to generate our model.yaml doc from the validator's source code.

For brevity of the draft I only generate the doc for "format_version", "language" and "source".
This is the markdown output (only minor formatting changes to what we have in our manual .md file now):

----------------------
* `format_version` Version of this BioImage.IO Model Description File Specification. This is mandatory, and important for the consumer software to verify before parsing the fields. The recommended behavior for the implementation is to keep backward compatibility, and throw error if the model yaml is in an unsupported format version. Current format version: 0.3.0
* `language` [optional]* Programming language of the source code. For now, we support python and java. This field is only required if the field source is present.
* `source` [optional]* Language and framework specific implementation. As some weights contain the model architecture, the source is optional depending on the present weight formats. `source` can either point to a local implementation: `<relative path to file>:<identifier of implementation within the source file>` or the implementation in an available dependency: `<root-dependency>.<sub-dependency>.<identifier>`.
For example: `./my_function:MyImplementation` or `core_library.some_module.some_function`
----------------------

the core idea is to add free text to the marshmallow fields:
```python
class Spec(PyBioSchema):
    format_version = fields.String(
        validate=validate.OneOf(raw_nodes.FormatVersion.__args__),
        required=True,
        bioimageio_doc="Version of this BioImage.IO Model Description File Specification. This is mandatory, and "
        "important for the consumer software to verify before parsing the fields. The recommended behavior for the "
        "implementation is to keep backward compatibility, and throw error if the model yaml is in an unsupported "
        f"format version. Current format version: {raw_nodes.FormatVersion.__args__[-1]}",
    )
```
This way validation (code) and documentation live very close together. 

side note: the "[optional]" is inserted manually by reading out the `field.required` attribute. The asterisk (indicating dependent optionality) is not automated, but expected to appear in the doc string for the field. Automating this and the very flexible (dynamic) validation steps seems to much overhead.


FynnBe:
If we continue with this, we might consider splitting off the `pybio.spec` module to live in https://github.com/bioimage-io/spec-bioimage-io  instead.

constantinpape:
I think this looks good, but we should first think about the more fundamental question of how we want to express the spec, see https://github.com/bioimage-io/spec-bioimage-io/issues/74#issuecomment-827476146.

FynnBe:
before advocating the marshmallow solution and this documentation generation further, I wanted to make sure this approach holds for the nested fields...
It went a bit out of hand and now the doc generation also includes the following features:
- sorting: optional fields after required ones
- sorting: alphabetically with option for manual override (atm only used to bring `format_version` to the top)
-  default description: schema/field class name

the fully generated doc is [here](https://github.com/bioimage-io/python-bioimage-io/blob/191e792c080bb03d119b8c93e3e0dddbfc3a716a/scripts/bioimageio_model_spec.md) (most fields have only the default description as I skipped copying over all our descriptions from the current docs for now).

FynnBe:
beside all the doc generation, this PR adds the optional fields `packaged_by` and `parent` to the validator as they were described in the [current docs at this time](https://github.com/bioimage-io/spec-bioimage-io/blob/36af72b6f5c2f5e5aeeb153bca983dd70e5b74f9/README.md), but missing in the validator (single source of truth for the win!)

generated/model_spec_0.3.0.json
```diff
@@ -29,6 +29,67 @@
             ],
             "type": "object"
         },
+        "ImplicitInputShape": {
```
constantinpape, line 32: I think we can get rid of the auto-generated json schema for now.
As far as I remember our recent discussions we probably don't need it; and so removing it would get rid of another potential source of confusion. (And adding it again in case we need it is cheap)

constantinpape (commented):
I think we can remove the auto-generated jsonschema for now.
Looks good otherwise!

PR #64, opened 2021-05-05
Implement prototype for spec builder helper function

constantinpape:


pybio/spec/utils/build_spec.py
```diff
@@ -0,0 +1,179 @@
...
+# utility functions to build the spec from python
+#
+
+
+# TODO refactor common stuff here
+def _build_spec():
+    pass
+
+
+# TODO this should handle both local file paths and urls and download the url somwhere temp
```
FynnBe, line 18: shouldn't it leave urls as is if this is going to become a model.yaml?

pybio/spec/utils/build_spec.py
```diff
@@ -0,0 +1,179 @@
...
+    git_repo=None,
+    attachments=None,
+    packaged_by=None,
+    parent=None,
+    run_mode=None,
+    config=None
+):
+    """
+    """
+    model = raw_nodes.Model
```
FynnBe, line 63: I think model should be a raw_nodes.Model instance...  (same for all other raw_nodes)

pybio/spec/utils/build_spec.py
constantinpape: It needs to download the urls in order to compute the hashes etc. The urls that are written to the model are indeed not changed.

pybio/spec/utils/build_spec.py
constantinpape: makes sense, I will fix it.

FynnBe:
serialization can be done like in (this round trip test](https://github.com/bioimage-io/python-bioimage-io/blob/db47c4a8edb629259213db287f46b208cd8d63a7/tests/test_pybio/test_spec/test_dump_spec.py#L8)

As discussed, a round trip would be good to have somewhere under the hood anyway to obtain the defaults that are only specified for deserialization.

constantinpape:
@FynnBe I just had a look at the example [here](https://github.com/bioimage-io/python-bioimage-io/blob/db47c4a8edb629259213db287f46b208cd8d63a7/tests/test_pybio/test_spec/test_dump_spec.py#L14) and it looks a bit non-conical.
Usally the python syntax is
```python
something.dump(file_stream)  # dump the object to the file opened with file_stream
serialized = something.dumps()  # return the serialized object in memory
```
Is this something marshmallow specific or can we change this?

FynnBe:
> Is this something marshmallow specific or can we change this?

This is indeed marshmallow specific: https://marshmallow.readthedocs.io/en/stable/api_reference.html


dump(obj,&nbsp;*[,&nbsp;many]) | Serialize an object to native Python data types according to this Schema’s fields.
-- | --
dumps(obj,&nbsp;*args[,&nbsp;many]) | Same as&nbsp;dump(), except return a JSON-encoded string.

constantinpape:
Ok, I merged master now and started with a test. However, dumping the raw_nodes [here](https://github.com/bioimage-io/python-bioimage-io/blob/spec-builder/tests/test_specs/test_utils/test_build_spec.py#L35) fails with:
```
self = <fields.ImportableSource(default=<marshmallow.missing>, attribute=None, validate=None, required=False, load_only=False... 'validator_failed': 'Invalid value.', 'invalid': 'Not a valid string.', 'invalid_utf8': 'Not a valid utf-8 string.'})>
value = 'bioimageio.sklearn.models.RandomForestClassifier', attr = 'source'
obj = Model(source='bioimageio.sklearn.models.RandomForestClassifier', sha256=None, kwargs={'n_estimators': 10, 'criterion':...5f9d49cc9c19a95b4b0adfdd7', authors=None, attachments=None, parent=None, opset_version=None, tensorflow_version=None)})
kwargs = {}

    def _serialize(self, value, attr, obj, **kwargs) -> typing.Optional[str]:
        if isinstance(value, raw_nodes.ImportableModule):
            return value.module_name + "." + value.callable_name
        elif isinstance(value, raw_nodes.ImportablePath):
            return value.filepath.as_posix() + "::" + value.callable_name
        else:
>           raise TypeError(f"{value} has unexpected type {type(value)}")
E           TypeError: bioimageio.sklearn.models.RandomForestClassifier has unexpected type <class 'str'>

../bioimageio/spec/fields.py:230: TypeError
```
So it looks like this is expecting the class already?! Any ideas how to fix this, @FynnBe?

FynnBe:
> So it looks like this is expecting the class already?! Any ideas how to fix this, @FynnBe?

Indeed. the weight ui is expected to be a `URI` as defined here: https://github.com/bioimage-io/python-bioimage-io/blob/92e12ee1395ed4c3bf8b99c918dffea74d5c57ce/bioimageio/spec/raw_nodes.py#L79

constantinpape:
Ok, this brings me to the following error:
```
test_specs/test_utils/test_build_spec.py:35: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../software/conda/miniconda3/envs/torch-em-test/lib/python3.8/site-packages/marshmallow/schema.py:556: in dump
    result = self._serialize(processed_obj, many=many)
../../../software/conda/miniconda3/envs/torch-em-test/lib/python3.8/site-packages/marshmallow/schema.py:520: in _serialize
    value = field_obj.serialize(attr_name, obj, accessor=self.get_attribute)
../../../software/conda/miniconda3/envs/torch-em-test/lib/python3.8/site-packages/marshmallow/fields.py:317: in serialize
    return self._serialize(value, attr, obj, **kwargs)
../bioimageio/spec/fields.py:230: in _serialize
    raise TypeError(f"{value} has unexpected type {type(value)}")
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = URI(scheme='bioimageio.sklearn.models.RandomForestClassifier', authority=<marshmallow.missing>, path=<marshmallow.missing>, query=<marshmallow.missing>, fragment=<marshmallow.missing>)

    def __str__(self):
        """scheme:[//authority]path[?query][#fragment]"""
        return (
>           (self.scheme + ":" if self.scheme else "")
            + ("//" + self.authority if self.authority else "")
            + self.path
            + ("?" + self.query if self.query else "")
            + ("#" + self.fragment if self.fragment else "")
        )
E       TypeError: can only concatenate str (not "_Missing") to str

../bioimageio/spec/raw_nodes.py:91: TypeError
```
It looks like the value for source here `bioimageio.sklearn.models.RandomForestClassifier` does not adhere to the URI schema.

constantinpape:
This is working now. See https://github.com/constantinpape/torch-em/blob/main/torch_em/util/modelzoo.py#L478 for example usage and https://github.com/constantinpape/torch-em/blob/main/test/util/test_modelzoo.py for a full test.
(We can maybe think about pulling some more functionality into the repo here).

@FynnBe there are still some TODOs left, especially w.r.t. other weight formats, but for these there might be some functionality in here already that can be reused.

bioimageio/spec/utils/build_spec.py
```diff
@@ -0,0 +1,246 @@
...
+    halo: Optional[List[int]] = None,
+    preprocessing: Optional[List[Dict[str, Dict[str, Union[int, float, str]]]]] = None,
+    postprocessing: Optional[List[Dict[str, Dict[str, Union[int, float, str]]]]] = None,
+    # general optional
+    git_repo: Optional[str] = None,
+    attachments: Optional[List[str]] = None,
+    packaged_by: Optional[List[str]] = None,
+    run_mode: Optional[str] = None,
+    parent: Optional[str] = None,
+    config=None  # TODO not sure what are the correct type hints for config
```
FynnBe, line 179:
```suggestion
    config: Optional[Dict[str, Any]] = None,
```

bioimageio/spec/utils/build_spec.py
```diff
@@ -0,0 +1,246 @@
+import os
+import datetime
+import hashlib
+from typing import Dict, List, Optional, Union
```
FynnBe, line 4:
```suggestion
from typing import Any, Dict, List, Optional, Union
```

if type hint suggestion below is accepted...

bioimageio/spec/utils/build_spec.py
```diff
@@ -0,0 +1,246 @@
...
+    test_inputs: List[str],
+    test_outputs: List[str],
+    # general required
+    name: str,
+    description: str,
+    authors: List[str],
+    tags: List[str],
+    license: str,
+    documentation: str,
+    covers: str,
```
FynnBe, line 154:
```suggestion
    covers: List[str],
```

raw_nodes.Model expects `List[URI]`, so there is an additional step required here...

bioimageio/spec/utils/build_spec.py
```diff
@@ -0,0 +1,246 @@
...
+        sha256=source_hash,
+        kwargs=model_kwargs,
+        format_version=format_version,
+        name=name,
+        description=description,
+        authors=authors,
+        cite=cite,
+        tags=tags,
+        license=license,
+        documentation=documentation,
```
FynnBe, line 233: should be `URI`, not `str`

bioimageio/spec/utils/build_spec.py
```diff
@@ -0,0 +1,246 @@
...
+    # general required
+    name: str,
+    description: str,
+    authors: List[str],
+    tags: List[str],
+    license: str,
+    documentation: str,
+    covers: str,
+    dependencies: str,
+    cite: Optional[List[str]],
```
FynnBe, line 156: looking at `_build_cite` it should probably be:
```suggestion
    cite: Dict[str, str],
```

bioimageio/spec/utils/build_spec.py
```diff
@@ -0,0 +1,246 @@
...
+    output_name: Optional[str] = None,
+    output_reference: Optional[str] = None,
+    output_scale: Optional[List[int]] = None,
+    output_offset: Optional[List[int]] = None,
+    halo: Optional[List[int]] = None,
+    preprocessing: Optional[List[Dict[str, Dict[str, Union[int, float, str]]]]] = None,
+    postprocessing: Optional[List[Dict[str, Dict[str, Union[int, float, str]]]]] = None,
+    # general optional
+    git_repo: Optional[str] = None,
+    attachments: Optional[List[str]] = None,
```
FynnBe, line 175: spec-bioimage-io: 
> Dictionary of text keys...

```suggestion
    attachments: Optional[Dict[str, Any]] = None,
```

FynnBe (changes requested):
not sure if for all the with `Optional` annoted kwargs `None` is actually a valid input, e.g. for `cite` it is not.  All other's might be filtered out, I haven't checked this in detail yet. 

This raises a bit more general question, that is if we do want to allow some null values and convert them to an empty list, dict, etc... as of now (and afaik) all types that can be empty do not allow a null value.

bioimageio/spec/utils/build_spec.py
constantinpape: > raw_nodes.Model expects `List[URI]`, so there is an additional step required here...

Some of these are apparently converted implicitly; passing just string here does work, it ends up being serialized correctly in the yaml.

bioimageio/spec/utils/build_spec.py
constantinpape: Yes, indeed. I had it with a default value before and forgot to remove the optional.

bioimageio/spec/utils/build_spec.py
constantinpape: To be completely precise it says:

> attachments [optional] Dictionary of text keys and URI (or a list of URI) 

So both dict and list. But I am not really sure why we would need a dict here; a flat list of files / URIs should be enough.

bioimageio/spec/utils/build_spec.py
constantinpape: Seems to work implicitly.

bioimageio/spec/utils/build_spec.py
FynnBe: oh, I read that as a dict of text keys and [URI (or a list of URI)]...
and that's what I have implemented in the validator
```python
    attachments = fields.Dict(
        fields.String,
        fields.Union([fields.URI(), fields.List(fields.URI)]),
        ...
        ```
I think the idea was to give the URIs some explicit name/tag/mini documentation and @oeway had suggested some special keys, e.g. "files" (or smth similar) to be downloaded and included by the packager...

Comment (constantinpape) [reply bioimageio/spec/utils/build_spec.py:175]: Ok, that makes sense :). Then it should be `Optional[Dict[str, Union[str, List[str]]]` though.

File: bioimageio/spec/utils/build_spec.py (hunk 6 above)
Comment (constantinpape) [reply bioimageio/spec/utils/build_spec.py:233]: > Maybe the best solution would be to include a round-trip in this function.

Yes, that would make sense. What's the best function for this though?
Ideally we would need something that does only formal verification, i.e. are the URIs well formed etc. but does not check if they actually exist etc. Otherwise we limit how this function can be used by forcing the user that all the data is in place already.

File: bioimageio/spec/utils/build_spec.py [hunk 9]
```diff
@@ -0,0 +1,246 @@
...
+        dependencies=dependencies,
+        timestamp=timestamp,
+        weights=weights,
+        inputs=[inputs],
+        outputs=[outputs],
+        test_inputs=test_inputs,
+        test_outputs=test_outputs,
+        **kwargs
+    )
+    return model
```
FynnBe, line 246:
```suggestion
    # round trip for validation and correct types
    serialized = schema.Model().dump(model)
    model = schema.Model().load(serialized)

    return model
```

bioimageio/spec/utils/build_spec.py
FynnBe: I thought of a round trip serialization/deserialization like so: https://github.com/bioimage-io/python-bioimage-io/pull/64/files/ce10c3bf1b59e80d2f07e4e876e3420c951b95e5#r639518930

FynnBe:
> because its input data does not have a batch dimension:

but it does: https://github.com/bioimage-io/python-bioimage-io/blob/7b9d4f9b5235262e886efab03d673bde05f5a370/specs/models/sklearn/RandomForestClassifier.model.yaml#L23

bioimageio/spec/utils/build_spec.py
```diff
@@ -0,0 +1,251 @@
...
+    """
+    #
+    # generate the model specific fields
+    #
+
+    # check the test inputs and auto-generate input/output description from test inputs/outputs
+    for test_in, test_out in zip(test_inputs, test_outputs):
+        test_in, test_out = _ensure_uri(test_in, root), _ensure_uri(test_out, root)
+        test_in, test_out = np.load(test_in), np.load(test_out)
+    inputs = _get_input_tensor(test_in, input_name, input_step, input_min_shape, preprocessing)
```
FynnBe, line 190: when running the test I get `inputs.axes = "bcyx"` and `input.shape = (10000, 1)`... so something is off

constantinpape: Yes, I know what's going on, I am fixing it now.

FynnBe: feel free to merge in https://github.com/bioimage-io/python-bioimage-io/pull/66 at any time that suits you. I just realized that the test dir was a bit out of order...

constantinpape:
> but it does:

Indeed, the problem was that `build_spec` did not expect 2d input data.

constantinpape:
I think all the issues from code review are fixed now, but there are still some TODOs where we can maybe use existing functionality:
- [Download URLs](https://github.com/bioimage-io/python-bioimage-io/blob/spec-builder/bioimageio/spec/utils/build_spec.py#L14) to generate the hash. @FynnBe can we reuse something in the repo here?
- [Support for more weight formats and format auto-detection](https://github.com/bioimage-io/python-bioimage-io/blob/spec-builder/bioimageio/spec/utils/build_spec.py#L31-L36)
- [Select doi or url for citation entry based on pattern matching](https://github.com/bioimage-io/python-bioimage-io/blob/spec-builder/bioimageio/spec/utils/build_spec.py#L162).

FynnBe:
> I think all the issues from code review are fixed now, but there are still some TODOs where we can maybe use existing functionality:
> 
> * [Download URLs](https://github.com/bioimage-io/python-bioimage-io/blob/spec-builder/bioimageio/spec/utils/build_spec.py#L14) to generate the hash. @FynnBe can we reuse something in the repo here?

maybe https://github.com/bioimage-io/python-bioimage-io/blob/5dae0d7a0e44a0005c197f1da00fea5474ecd849/bioimageio/spec/utils/transformers.py#L295
but there nothing to compute the hash yet
> * [Support for more weight formats and format auto-detection](https://github.com/bioimage-io/python-bioimage-io/blob/spec-builder/bioimageio/spec/utils/build_spec.py#L31-L36)
there is not weight auto-detection yet, as we were just validating... The defined weight formats are [here](https://github.com/bioimage-io/python-bioimage-io/blob/5dae0d7a0e44a0005c197f1da00fea5474ecd849/bioimageio/spec/raw_nodes.py#L30)
> * [Select doi or url for citation entry based on pattern matching](https://github.com/bioimage-io/python-bioimage-io/blob/spec-builder/bioimageio/spec/utils/build_spec.py#L162).
unfortunately doi and url entries in the `CiteEntry` schema are still very simple (no checks at all): https://github.com/bioimage-io/python-bioimage-io/blob/5dae0d7a0e44a0005c197f1da00fea5474ecd849/bioimageio/spec/schema.py#L30

I think we should improve the `schema.CiteEntry` and then just try except that schema  with doi and url...

constantinpape:
> maybe
> 
> https://github.com/bioimage-io/python-bioimage-io/blob/5dae0d7a0e44a0005c197f1da00fea5474ecd849/bioimageio/spec/utils/transformers.py#L295

Cool, I will check it out. I am calculating the hash already anyways as this is just one hashlib call.



> there is no weight auto-detection yet, as we were just validating... The defined weight formats are here:
> 
> https://github.com/bioimage-io/python-bioimage-io/blob/5dae0d7a0e44a0005c197f1da00fea5474ecd849/bioimageio/spec/raw_nodes.py#L30-L38

Yes, I was just looking for that list, I can implement the auto detection here.

> I think we should improve the `schema.CiteEntry` and then just try except that schema with doi and url...

Yes, that would be very helpful.

constantinpape:
I updated `build_spec` to also support additional weight formats and also updated the test. However, the tests for onnx and torchscript fail. I think it's because `source = None`:
```
self = <fields.ImportableSource(default=<marshmallow.missing>, attribute=None, validate=None, required=False, load_only=False... 'validator_failed': 'Invalid value.', 'invalid': 'Not a valid string.', 'invalid_utf8': 'Not a valid utf-8 string.'})>
value = None, attr = 'source'
obj = Model(source=None, sha256=None, kwargs={'input_channels': 1, 'output_channels': 1}, format_version='0.3.1', name='UNet...issing>, parent=<marshmallow.missing>, opset_version=<marshmallow.missing>, tensorflow_version=<marshmallow.missing>)})
kwargs = {}

    def _serialize(self, value, attr, obj, **kwargs) -> typing.Optional[str]:
        if isinstance(value, raw_nodes.ImportableModule):
            return value.module_name + "." + value.callable_name
        elif isinstance(value, raw_nodes.ImportablePath):
            return value.filepath.as_posix() + "::" + value.callable_name
        else:
>           raise TypeError(f"{value} has unexpected type {type(value)}")
E           TypeError: None has unexpected type <class 'NoneType'>
```
But according to the spec, the source should be optional (and indeed it's not necessary for onnx or torchscript).
Could you check this @FynnBe?

constantinpape:
@FynnBe thanks for fixing the issue with the source etc.
I think this PR is good to go now!

FynnBe (approved):
yeah, LGTM!

PR #65, opened 2021-05-07
Refactor package name

FynnBe:
mostly straight forward refactoring of the package name.

the rf dummy model needed new pickle weights, which is fine, as pickle is not officially supported anyway, but torch script weights might be affected similarly, see failing tests in https://github.com/ilastik/tiktorch/pull/164


constantinpape (approved):
Looks good to me. Feel free to merge when you think this is all ready.

PR #73, opened 2021-06-10
Clean up specs and tests

constantinpape:
I noticed that some of the specs and tests are still duplicate.
@FynnBe when I looked into this, I was a bit confused that the manifest.yaml file is empty and seems to have been empty all along. Was that intentional or did we just forget to add it.


FynnBe:
the pickle based RF model is inofficial, it should never find its way to bioimage.io. Therefore it was not in the manifest and there are/were no other models in this repo...

FynnBe:
removing it here to keep it only in spec-bioimage-io does make sense to me (now ;-))

constantinpape:
> removing it here to keep it only in spec-bioimage-io does make sense to me (now ;-))

Ok.

Also as far as I can tell everything here: https://github.com/bioimage-io/python-bioimage-io/tree/master/bioimageio
is deprecated and can just be removed. Or is there anything we still need to keep, @FynnBe?

FynnBe:
we should keep transformations and protocols in [core](https://github.com/bioimage-io/python-bioimage-io/tree/master/bioimageio/core) they'll need an update but are not completely deprecated

FynnBe:
to some extend the same applies for [runners](https://github.com/bioimage-io/python-bioimage-io/tree/master/bioimageio/runners) which is basically a draft.

FynnBe:
oh and [sklearn](https://github.com/bioimage-io/python-bioimage-io/tree/master/bioimageio/sklearn) is used in the testing RF model

FynnBe:
we might want to absorb `runners` into the core module when overhauling it

constantinpape:
> we should keep transformations and protocols in [core](https://github.com/bioimage-io/python-bioimage-io/tree/master/bioimageio/core) they'll need an update but are not completely deprecated

Hmm, isn't this all part of the first spec attempt and deprecated? Looking at [transformations](https://github.com/bioimage-io/python-bioimage-io/tree/master/bioimageio/core/transformations) I don't think this has anything to do with the current pre/postprocessing logic.

> we might want to absorb `runners` into the core module when overhauling it

I thought we want to abandon this in favor of the `PredictionPipeline` that will be moved here from tiktorch.

> oh and [sklearn](https://github.com/bioimage-io/python-bioimage-io/tree/master/bioimageio/sklearn) is used in the testing RF model

Ok, we can leave it in until we remove the pickle stuff from the spec.

FynnBe:
hmm.. with all the major updates and renaming coming in we might as well delete it in-between 👍

bioimageio/weight_converter/torch/torchscript.py
```diff
@@ -65,9 +65,9 @@ def main():
     parser.add_argument("--tracing", "-t", default=1, type=int)
 
     args = parser.parse_args()
-    return convert_weights_to_torchscript(os.path.abspath(args.model),
-                                          args.output,
-                                          bool(args.tracing))
+    return convert_weights_to_pytorch_script(os.path.abspath(args.model),
```
oeway, line 68: A quick jump-in comment, isn't `TorchScript ` a fixed term?

constantinpape: Yeah, I was also wondering about this, but we call the weight format `pytorch_script`, so I wanted to make it consistent here.

oeway: Aha, so it might be better to rename the format instead?

constantinpape: We could set it onto the list of potential 0.4 changes.

constantinpape:
@FynnBe I continued here now:
- cleanup: remove everything from core (all deprecated imo)
- weight_converter: moved weight converter functionality here (for now torch->torchscript, torch->onnx)

setup.py
```diff
@@ -4,10 +4,11 @@
 # Get the long description from the README file
 long_description = (Path(__file__).parent / "README.md").read_text(encoding="utf-8")
 
+# TODO how do we deal with different sub-packages here?
 setup(
-    name="bioimageio.core",
+    name="bioimageio.weight_converter",
```
FynnBe, line 9: I suppose we should think about what bioimageio packages we want to end up with and what functionality we can maybe group together... but we can totally go ahead and have thie bioimageio.weight_converter package for now and change name/content later if need be...

FynnBe (approved):
LGTM

setup.py
constantinpape: > I suppose we should think about what bioimageio packages we want to end up with and what functionality we can maybe group together... but we can totally go ahead and have thie bioimageio.weight_converter package for now and change name/content later if need be...

I agree, we need some proper discussion on how to structure this repo, see https://github.com/bioimage-io/python-bioimage-io/issues/74. But moving/renaming the `weight_converter` things is easy if we decide to, so I will go ahead and merge.

PR #76, opened 2021-07-23
Add build spec

constantinpape:
Moves over the `build_spec` functionality from bioimage.io

TODO:
- [x] fix the test (need to create a pytest config)
- [ ] decide on naming for the module and functions; would probably be good to be more specific

Depends on #75 

cc @FynnBe @k-dominik

### DIFF
--- a/bioimageio/core/build_spec/build_spec.py
+++ b/bioimageio/core/build_spec/build_spec.py
@@ -1,0 +1,447 @@
+import datetime
+import hashlib
+import os
+from pathlib import Path
+from typing import Any, Dict, List, Optional, Union
+
+import numpy as np
+from ruamel.yaml import YAML
+
+import bioimageio.spec as spec
+import bioimageio.spec.model as model_spec
+
+try:
+    from typing import get_args
+except ImportError:
+    from typing_extensions import get_args  # type: ignore
+
+#
+# utility functions to build the spec from python
+#
+
+
+def _get_local_path(uri, root=None):
+    is_local_path = os.path.exists(uri)
+    if not is_local_path and root is not None:
+        uri2 = os.path.join(root, uri)
+        if os.path.exists(uri2):
+            uri = uri2
+            is_local_path = True
+    if not is_local_path:
+        uri = spec.shared.fields.URI().deserialize(uri)
+        uri = spec.shared.download_uri_to_local_path(uri).as_posix()
+    return uri
+
+
+def _get_hash(path):
+    with open(path, "rb") as f:
+        data = f.read()
+        return hashlib.sha256(data).hexdigest()
+
+
+def _infer_weight_type(path):
+    ext = os.path.splitext(path)[-1]
+    if ext in (".pt", ".torch"):
+        return "pytorch_state_dict"
+    elif ext == ".onnx":
+        return "onnx"
+    elif ext in (".hdf", ".hdf5", ".h5"):
+        return "keras_hdf5"
+    elif ext == ".zip":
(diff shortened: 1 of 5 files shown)


constantinpape:
I fixed most tests, but I am not sure how to do the for `pytorch_state_dict`, which needs the `source`; and which must be a local file rn.

bioimageio/core/build_spec/build_spec.py
```diff
@@ -0,0 +1,447 @@
...
+    from typing import get_args
+except ImportError:
+    from typing_extensions import get_args  # type: ignore
+
+#
+# utility functions to build the spec from python
+#
+
+
+def _get_local_path(uri, root=None):
```
FynnBe, line 23: this could actually be substituted with `bioimageio.spec.shared.utils.resolve_uri`

constantinpape:
@FynnBe I am using `resolve_uri` now and fixed the issue in the test. So in principle this is good to go; but we could still think about more descriptive names for the `build_spec` submodule and its functions.

bioimageio/core/build_spec/build_spec.py
```diff
@@ -0,0 +1,440 @@
...
+
+
+# TODO The citation entry should be improved so that we can properly derive doi vs. url
+def _build_cite(cite: Dict[str, str]):
+    citation_list = [model_spec.raw_nodes.CiteEntry(text=k, url=v) for k, v in cite.items()]
+    return citation_list
+
+
+# TODO we should make the name more specific: "build_model_spec"?
+# TODO maybe "build_raw_model" as it return raw_nodes.Model
```
FynnBe, line 223: I think 'build_raw_model' is a good idea, as we might add `build_model` in the future (for now that does not make so much sense as there is no serialization from non-raw nodes...

bioimageio/core/build_spec/build_spec.py
```diff
@@ -0,0 +1,440 @@
...
+    new_weights = _get_weights(weight_uri, weight_type, None, root, **weight_kwargs)[0]
+    model.weights.update(new_weights)
+
+    serialized = model_spec.schema.Model().dump(model)
+    model = model_spec.schema.Model().load(serialized)
+
+    return model
+
+
+def serialize_spec(model, out_path):  # TODO change name to include model (see build_model_spec)
```
FynnBe, line 436: we can substitute this as well with `bioimageio.spec.save_raw_node`

FynnBe (approved):
Already LGTM. name changes could occur later as well...
maybe just substitute `serialize_spec` before merging.

bioimageio/core/build_spec/build_spec.py
constantinpape: Ok, I will check if that works.

bioimageio/core/build_spec/build_spec.py
constantinpape: I decided to go with `build_model` here. I find the `raw` things very confusing in the function names.

bioimageio/core/build_spec/build_spec.py
constantinpape: removed

constantinpape:
@FynnBe since I updated the biomageio.spec to the latest main the test imports don't work any more:
```
test_build_spec.py:4: in <module>
    from bioimageio.spec.model import schema
E   ModuleNotFoundError: No module named 'bioimageio'
```

Any idea what's causing this?

constantinpape:
Looks like the import issue is just something wrong with my dev env on the cluster; works locally, so I'm merging this.

PR #82, opened 2021-07-26
Upd conda recipe

k-dominik:
Summary:

* Added a "first" conda recipe. setup.py is currently still source of truth for dependencies/constraints. Used run_constrained section to specify versions for optional packages.
* Added some installation instructions
* Added basic ci

Todo:

* [ ] come up with sensible minimal (max) requirements for onnxruntime, tensorflow (are there additional dependencies that go with those?)
* [x] add ci for various configs

supersedes #79


k-dominik:
hmm I guess this package relies on an unreleased version of `bioimageio.spec` (which seems not to be compatible with the latest current release). Of course we could start relying on the main branch of `bioimageio.spec` or adding it as a submodule. Distributing `bioimageio.core` independently doesn't make a lot of sense then (imo), with outdated upstream dependency (`bioimageio.spec`).

CC: @FynnBe @constantinpape

constantinpape:
> Edit: Alternatively we could just add more "configurations" here. Once using `bioimageio.spec` `main`, and once with the latest package.

Yes, I think that would be my preferred solution.

k-dominik:
alright, this can be reviewed - note:

test failures against conda `bioimage.spec` package are expected. Added tests against `main`.
Current conda recipe will of course not work, as `bioimageio.spec` is outdated on conda-forge.

README.md
```diff
@@ -1,6 +1,37 @@
...
-# Model Specification
-The model specification and its validation tools have moved to https://github.com/bioimage-io/spec-bioimage-io.
+## Installation
+
+The `bioimageio.core` package supports various back-ends for running BioimageIO networks:
+
+* Pytorch/Torchscript:
+  ```bash
+  # cpu installation (if you don't have an nvidia graphics card)
+  conda install -c pytorch -c conda-forge -c ilastik-forge bioimageio.core pytorch torchvision cpuonly
```
constantinpape, line 12: Why do we need `ilastik-forge` in the channels here? Same for the conda install commands below.

dev/environment-base.yaml
```diff
@@ -0,0 +1,12 @@
+name: bio-core-dev
+channels:
+  - pytorch
+  - conda-forge
+  - defaults
+dependencies:
```
constantinpape, line 6: Should we add pytorch in the dependencies? Or remove the pytorch channel?

constantinpape (commented):
Some minor comments, looks good otherwise.
I would vote for merging this despite the failing tests, so that we can check that things work with spec main.

README.md
k-dominik: Not really once we have the package on conda-forge. Until then we could have some versions on ilastik-forge. Not a very strong opinion there. But this makes it easier to remove certain versions in the beginning. (I of course know of such mechanisms in the conda-forge community, but would not want to put too much stress on the community. I'd prefer to come to a somewhat stable api first).

dev/environment-base.yaml
k-dominik: right, in principle we should test

* no NN backend
* Pytorch
* onnxruntime
* tensorflow2
?

k-dominik:
> Some minor comments, looks good otherwise.
> I would vote for merging this despite the failing tests, so that we can check that things work with spec main.

okay I think we can potentially merge this. This will test with a single (for each python version) environment (cpuonly) with `onnxruntime`, `pytorch`, `tensorflow`.

I guess we can add pins/minimal versions to those later when needed.

constantinpape:
Thanks Dominik!

PR #84, opened 2021-08-02
add version file and use in setup.py

FynnBe:
copied the version file pattern from spec-bioimage-io


constantinpape (approved):
Looks good and test failures look unrelated to the changes here.
I need this in #92, so I will merge it now.

PR #86, opened 2021-08-03
Add tests for prediction pipeline; add keras model adapter

constantinpape:
I have started to add the keras model adapter (not tested yet) and tests for the prediction pipeline.
One thing I noticed is that the device management in the prediction pipeline seems to be tailored to torch right now.
We should think about more framework agnostic device management.

Also, the test for the torchscript model currently fails with `E           RuntimeError: [enforce fail at inline_container.cc:145] . PytorchStreamReader failed reading zip archive: unsupported multidisk archive   `.
It seems like the torchscript weights in the example model are corrupted....

### DIFF
--- a/bioimageio/core/predict.py
+++ b/bioimageio/core/predict.py
@@ -2,8 +2,6 @@ import argparse
 import sys
 
 import numpy as np
-# FIXME this is too torch specific
-import torch.cuda
 import xarray as xr
 
 from bioimageio.core.prediction_pipeline import create_prediction_pipeline
@@ -14,6 +12,7 @@ parser = argparse.ArgumentParser()
 parser.add_argument("-m", "--model", help="bioimage model resource (ziped package or rdf.yaml)", required=True)
 parser.add_argument("images", nargs="+", dest="image(s)", help="image(s) to process (.npy)")
 parser.add_argument("-o", "--outputs", nargs="+", dest="output(s)", help="output image(s) (.npy)", required=True)
+parser.add_argument("--devices", nargs="+", help="Devices to run this model", default=None)
 
 
 def main():
@@ -21,7 +20,7 @@ def main():
     model = load_resource_description(args.model)
     assert isinstance(model, Model)
     prediction_pipeline = create_prediction_pipeline(
-        bioimageio_model=model, devices=["cuda" if torch.cuda.is_available() else "cpu"]
+        bioimageio_model=model, devices=args.devices
     )
 
     if len(args.images) != len(model.inputs):
--- a/bioimageio/core/predict.py
+++ b/bioimageio/core/predict.py
@@ -19,9 +19,7 @@ def main():
     args = parser.parse_args()
     model = load_resource_description(args.model)
     assert isinstance(model, Model)
-    prediction_pipeline = create_prediction_pipeline(
-        bioimageio_model=model, devices=args.devices
-    )
+    prediction_pipeline = create_prediction_pipeline(bioimageio_model=model, devices=args.devices)
 
     if len(args.images) != len(model.inputs):
         raise ValueError(f"Expected {len(model.inputs)} input images, not {len(args.images)}")
@@ -30,8 +28,9 @@ def main():
         raise ValueError(f"Expected {len(model.outputs)} output images, not {len(args.outputs)}")
 
     input_tensors = [np.load(ipt) for ipt in args.images]
-    tagged_data = [xr.DataArray(ipt_tensor, dims=tuple(ipt.axes))
-                   for ipt_tensor, ipt in zip(input_tensors, model.inputs)]
+    tagged_data = [
+        xr.DataArray(ipt_tensor, dims=tuple(ipt.axes)) for ipt_tensor, ipt in zip(input_tensors, model.inputs)
+    ]
     if len(tagged_data) > 1:
         raise NotImplementedError(len(tagged_data))
 
--- a/bioimageio/core/predict.py
+++ b/bioimageio/core/predict.py
@@ -2,6 +2,7 @@ import argparse
 import sys
 
 import numpy as np
+# FIXME this is too torch specific
 import torch.cuda
 import xarray as xr
 
@@ -30,7 +31,8 @@ def main():
         raise ValueError(f"Expected {len(model.outputs)} output images, not {len(args.outputs)}")
 
     input_tensors = [np.load(ipt) for ipt in args.images]
-    tagged_data = [xr.DataArray(ipt_tensor, dims=tuple(ipt.axes)) for ipt_tensor, ipt in zip(input_tensors, model.inputs)]
+    tagged_data = [xr.DataArray(ipt_tensor, dims=tuple(ipt.axes))
+                   for ipt_tensor, ipt in zip(input_tensors, model.inputs)]
     if len(tagged_data) > 1:
         raise NotImplementedError(len(tagged_data))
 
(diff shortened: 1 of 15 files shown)


constantinpape:
@FynnBe the tf / keras tests fail with 

```
self = <bioimageio.spec.shared.utils.UriNodeTransformer object at 0x7f843fff2580>, leaf = PosixPath('/tmp/bioimageio_cache/auto-convert/-821963116692508699/auto_converted_documentation.md')                                                          
                                                                                                                                                                                                                                                       
    def _transform_Path(self, leaf: pathlib.Path):                                                                                                                                                                                                     
>       assert not leaf.is_absolute()                                                                                                                                                                                                                  
E       AssertionError 
```

So again some issue with the auto converted documentation...

FynnBe:
> @FynnBe the tf / keras tests fail with
> 
> ```
> self = <bioimageio.spec.shared.utils.UriNodeTransformer object at 0x7f843fff2580>, leaf = PosixPath('/tmp/bioimageio_cache/auto-convert/-821963116692508699/auto_converted_documentation.md')                                                          
>                                                                                                                                                                                                                                                        
>     def _transform_Path(self, leaf: pathlib.Path):                                                                                                                                                                                                     
> >       assert not leaf.is_absolute()                                                                                                                                                                                                                  
> E       AssertionError 
> ```
> 
> So again some issue with the auto converted documentation...
> 
> p.s. this is also happening for all the example zero-cost models we want to support. Is there a way to properly fix this @FynnBe?
> Otherwise I would strongly suggest that we drop the requirement for the local doc again, this is a huge pain!

should be fixed by https://github.com/bioimage-io/spec-bioimage-io/pull/218 
Can't reproduce atm (missing keras env...)

constantinpape:
Yes, that fixes the particular error, now it fails with 
```
self = Model(format_version='0.3.2', name='Small Extracellular Vesicle TEM Segmentation (Fully Residual U-Net)', type='model'...e/Small_Extracellular_Vesicle_TEM_Segmentation_(Fully_R
esidual_U-Net)p0_unzipped/tensorflow_saved_model_bundle.zip'))})                                                                                                                       
                                                                                                                                                                                       
    def __post_init__(self):                                                                                                                                                           
        from . import nodes, raw_nodes                                                                                                                                                 
                                                                                                                                                                                       
        if not isinstance(self, (nodes.Node, raw_nodes.RawNode)):                                                                                                                      
            raise TypeError("base nodes should not be instantiated!")                                                                                                                  
                                                                                                                                                                                       
        for f in dataclasses.fields(self):                                                                                                                                             
            if getattr(self, f.name) is missing and (                                                                                                                                  
                get_origin(f.type) is not Union or not isinstance(missing, get_args(f.type))                                                                                           
            ):                                                                                                                                                                         
>               raise TypeError(f"{self.__class__}.__init__() missing required argument: '{f.name}'")                                                                                  
E               TypeError: <class 'bioimageio.spec.model.v0_3.nodes.Model'>.__init__() missing required argument: 'source' 
```

FynnBe:
Yep. It's a missing "Union[Missing, ..." annotation in the raw model node. I'll fix it when I'm back later today

FynnBe:
Current error:
```
from bioimageio.core.utils import get_nn_instance
E   ModuleNotFoundError: No module named 'bioimageio.core.utils
```
get_nn_instance was moved to the pytorch model adapter...

constantinpape:
@FynnBe @k-dominik I have added the keras model adapter, added tests for the prediction pipeline and moved the device logic into the model adapters so that we don't rely on torch for this.

The tests for tf and keras are still failing because of issues with loading the weights; but I think we should go ahead and merge this anyway, because I think these issues are model specific and will need to be fixed together with @esgomezm.

bioimageio/core/prediction_pipeline/_model_adapters/_onnx_model_adapter.py
```diff
@@ -30,6 +30,7 @@ def __init__(self, *, bioimageio_model: nodes.Model, devices=List[str]):
         onnx_inputs = self._session.get_inputs()
         assert len(onnx_inputs) == 1, f"expected onnx model to have one input got {len(onnx_inputs)}"
         self._input_name = onnx_inputs[0].name
+        # TODO onnx device management
```
k-dominik, line 33: what happens with empty devices? Does it run on cpu? In any case maybe one should issue a warning here. (same goes for `TensorflowModelAdapterBase`).

tests/conftest.py
```diff
@@ -2,6 +2,28 @@
 from bioimageio.spec import export_resource_package
 
 
+# set 'skip_<FRAMEWORK>' flags as global pytest variables,
+# to deselect tests that require frameworks not available in current env
+def pytest_configure():
```
k-dominik, line 7: initially I thought "awesome", but now, I'm not sure if this is a good idea. First of all creating the development environment would ideally install all dependencies (imo) to run all tests. Then skipping some for missing deps is not necessary. The only issue might arise if there are conflicts that make it impossible to install everything into the same env.

bioimageio/core/prediction_pipeline/_model_adapters/_onnx_model_adapter.py
constantinpape: The empty device list is just a place holder until someone implements proper device management here... I think in both cases (tf and onnx) it runs on CPU. 
I agree, we should at least add a warning that arguments are ignored (if the devices are not None).

tests/conftest.py
constantinpape: It also depends if we want tests to run for users who might only set up an environment with a subset of the dependencies.
I think it would also be quite useful to add another "blank" env for testing that does not contain any of the frameworks, to check that the module can be imported, and that base functionality works, with missing frameworks.

bioimageio/core/prediction_pipeline/_model_adapters/_onnx_model_adapter.py
constantinpape: I added warnings for this.

FynnBe (approved):
some minor comments only

bioimageio/core/predict.py
```diff
@@ -30,7 +30,8 @@ def main():
         raise ValueError(f"Expected {len(model.outputs)} output images, not {len(args.outputs)}")
 
     input_tensors = [np.load(ipt) for ipt in args.images]
-    tagged_data = [xr.DataArray(ipt_tensor, dims=tuple(ipt.axes)) for ipt_tensor, ipt in zip(input_tensors, model.inputs)]
+    tagged_data = [xr.DataArray(ipt_tensor, dims=tuple(ipt.axes))
+                   for ipt_tensor, ipt in zip(input_tensors, model.inputs)]
```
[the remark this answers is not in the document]
constantinpape: I think the line was longer than 120.

tests/weight_converter/torch/test_onnx.py
```diff
@@ -1,23 +1,15 @@
...
-    import onnxruntime as rt
-except ImportError:
-    rt = None
-
-
-@pytest.mark.skipif(torch is None, reason="requires pytorch")
-@pytest.mark.skipif(rt is None, reason="requires onnx")
+@pytest.mark.skipif(pytest.skip_torch, reason="requires pytorch")
 def test_torchscript_converter(unet2d_nuclei_broad_model, tmp_path):
     from bioimageio.core.weight_converter.torch.onnx import convert_weights_to_onnx
```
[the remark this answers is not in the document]
constantinpape: No, the weight converter does not require onnx. It only has it as optional dependency to check the result, see https://github.com/bioimage-io/python-bioimage-io/blob/main/bioimageio/core/weight_converter/torch/onnx.py#L8-L11.

constantinpape:
I will go ahead and merge this.

PR #90, opened 2021-08-04
Update test-env

constantinpape:
Trying to find an env that's compatible with the deepimagej models ... WIP

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@ Python specific core utilities for working with the BioimageIO model zoo.
 
 ## Installation
 
+### Via Conda
+
 The `bioimageio.core` package supports various back-ends for running BioimageIO networks:
 
 * Pytorch/Torchscript:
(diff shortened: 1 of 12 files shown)


constantinpape:
@k-dominik @FynnBe I think this has working inference for keras and tensorflow, including a compatible env now. 
The tests for tf and keras are still failing, but I think this is due to genuine issues with the models. I will try to have a look at that with @esgomezm tomorrow. 

If you have time you could have a look at this already. Note that I added a hacky version of `scale_range` because it was needed for one of the deepimageJ models, before merging this we should have another look at that function and add a test.

FynnBe (approved):
LGTM

constantinpape:
The tests for keras and tf inference still fail due to issues with the example model; I have checked with a local model that this works in principle. I am merging this so that others can use the inference engine.

PR #91, opened 2021-08-05
Update build_model function

constantinpape:


constantinpape:
Only tests with known issues are failing, so I am merging.

PR #92, opened 2021-08-05
Add bulk prediction script

constantinpape:
I needed a script to predict all images in a folder; it might make sense to add it here, as the current functionality in the predict script is fairly limited.

@FynnBe @k-dominik do you think this is something we want to add here? I personally find this very useful, but I am not sure if you want it to be maintained here; I could also add it to my dl repo.


FynnBe:
IMO it makes sense to have it here. potentially by just expanding the predict script with some glob pattern... ?

constantinpape:
> predict script with some glob pattern... ?

Yes, I also thought about that. 

It's a bit unfortunate that there is so much axis logic that needs to be introduced, but I am not sure if there's anything we could reuse here.

constantinpape:
@FynnBe @k-dominik this is ready for review now. In this PR:
- Move all CLI functionality to `__main__.py`.
- Add different prediction modes:
  - single input
  - multiple input
  - with padding
  - with tiling

FynnBe:
could we make the test_cli tests pass before merging?

FynnBe (approved):
👍

constantinpape:
The CLI tests pass now; I replaced `conda develop` with `pip install -e`; looks like there is some issue with installing CLI otherwise.

PR #93, opened 2021-08-05
Move everything not purely "spec" from bioimageio.spec over here

FynnBe:


FynnBe:
tests fail due to frunet using `tiff` format for test input and output data, which we try to open with `numpy.load()`...
besides that everything should be up and running now.

constantinpape:
The tests fail with
```
 ImportError while loading conftest '/home/runner/work/python-bioimage-io/python-bioimage-io/tests/conftest.py'.
tests/conftest.py:2: in <module>
    from bioimageio.core import export_resource_package
bioimageio/core/__init__.py:4: in <module>
    from .resource_io import (
bioimageio/core/resource_io/__init__.py:1: in <module>
    from .io_ import (
bioimageio/core/resource_io/io_.py:12: in <module>
    from bioimageio.core.resource_io.nodes import ResourceDescription
bioimageio/core/resource_io/nodes.py:119: in <module>
    class _WeightsEntryBase(Node, model_raw_nodes._WeightsEntryBase):
E   AttributeError: module 'bioimageio.spec.model.v0_3.raw_nodes' has no attribute '_WeightsEntryBase'
```
so I guess we should first merge https://github.com/bioimage-io/spec-bioimage-io/pull/226?

FynnBe:
> so I guess we should first merge [bioimage-io/spec-bioimage-io#226](https://github.com/bioimage-io/spec-bioimage-io/pull/226)?

yes, indeed it requires those changes

FynnBe:
functionality moved over from bioimageio.spec:
- `export_resource_package` (create zip (model) package)
- `load_raw_resource_description` (extends `spec.load_raw_resource_description_from_dict` with excepting paths and urls as well)
- `load_resource_description` (loas raw resource description and applies various convenience functions. returns (non-raw) resource description)
- `ensure_raw_resource_description` (returns tuple of raw resource description and root_path. may be dropped in the future, if raw resources may know where their root lies... this needs further thought and discussion: )

note how deprecated format versions are not supported in bioimageio.core:
bioimageio.spec defines all (past) format versions for different resource types. therefore in bioimageio.spec some helper functions have an 'update_format' flag to selectively appy auto-conversion to the newest appropriate format. In bioimageio.core however we now work only with the latest format versions as defined in the bioimageio.spec module we depend on. If one wants to work with older format versions using the bioimageio.core package they should use a previously released bioimageio.core package. Multiple (deprecated) format versions in one bioimageio.core package are not supported unless the previous format version can be auto-converted to the 'latest'.

constantinpape:
Sounds good. See only my comment regarding `load_raw_resource_description` in https://github.com/bioimage-io/spec-bioimage-io/pull/226#issuecomment-914575277. (I don't have a strong feeling where this should go; if you think it should rather go into `core` than into `spec` that's also fine with me, I just want to double check because I think we discussed the opposite this afternoon.)

FynnBe:
I'm surprised this error: 
`attempted to read a PyTorch file with version 3, but the maximum supported version for reading is 2. Your PyTorch installation may be too old.`
hasn't showed up before...

FynnBe:
another error is that https://sandbox.zenodo.org/record/894498/files/rdf.yaml contains links to files that do not exist (attachments:files contains exampleImage.tif and resultImage.tif with .tif (single f), instead of .tiff). note that it should not be required to specify them in attachments:files as they are also present in test inputs/outputs and would thus be packaged anyway.

constantinpape:
> I'm surprised this error:
> `attempted to read a PyTorch file with version 3, but the maximum supported version for reading is 2. Your PyTorch installation may be too old.`
> hasn't showed up before...

In which test does this happen? Has pytorch changed dataformat again?!

> another error is that https://sandbox.zenodo.org/record/894498/files/rdf.yaml contains links to files that do not exist (attachments:files contains exampleImage.tif and resultImage.tif with .tif (single f), instead of .tiff). note that it should not be required to specify them in attachments:files as they are also present in test inputs/outputs and would thus be packaged anyway

Yes, I know. The test models for tf/keras are not in a good state. Eventually I would like to add example models for this to the spec repo, but we need to coordinate this with @esgomezm.

FynnBe:
> In which test does this happen? Has pytorch changed dataformat again?!

I think the test environment setup is taking a wrong turn...
```
...
The following packages will be SUPERSEDED by a higher-priority channel:

  pytorch            conda-forge::pytorch-1.9.0-cpu_py37hb~ --> pytorch::pytorch-1.4.0-py3.7_cuda10.1.243_cudnn7.6.3_0
  ```

constantinpape:
> > In which test does this happen? Has pytorch changed dataformat again?!
> 
> I think the test environment setup is taking a wrong turn...
> 
> ```
> ...
> The following packages will be SUPERSEDED by a higher-priority channel:
> 
>   pytorch            conda-forge::pytorch-1.9.0-cpu_py37hb~ --> pytorch::pytorch-1.4.0-py3.7_cuda10.1.243_cudnn7.6.3_0
> ```

Yes, that's wrong ... I guess something on conda changed that messes this up; could yo try making conda-forge channel higher priority than pytorch, looks like that would fix it.

FynnBe:
`test_onnx` now passes

constantinpape:
Yeah, let's go ahead and merge.

PR #97, opened 2021-09-09
Add more env files and update gh actions

constantinpape:


constantinpape:
@FynnBe, @k-dominik this introduces two new envs, one only with pytorch + onnx, the other only with tf 1.15 + keras and uses those in gh actions. This way we get around the dependency issues.

FynnBe (approved):
LGTM

PR #104, opened 2021-09-13
Multi tensor weight conversion

constantinpape:
Implement support for multiple input / output tensors in weight converter.
@FynnBe the changes in `conftest.py` are also relevant for #103. Also, I have added torchscript and onnx weights to the multi-tensor models, so you can test prediction with them.


FynnBe (approved):
LGTM

PR #105, opened 2021-09-15
Fixed shape support

FynnBe:


tests/conftest.py
```diff
@@ -2,75 +2,91 @@
...
-    if not pytest.skip_tensorflow:
-        load_packages |= set(keras_models)
-        load_packages |= set(tensorflow_js_models)
-        if pytest.tf_major_version == 1:
-            load_packages |= set(tensorflow1_models)
-        elif pytest.tf_major_version == 2:
-            load_packages |= set(tensorflow2_models)
-
-    pytest.model_packages = {name: export_resource_package(model_sources[name]) for name in load_packages}
+    pytest.skip_torch = skip_torch
```
constantinpape, line 80: Are you still using the skip variables at all?

tests/test_prediction.py
```diff
@@ -15,28 +15,30 @@ def test_test_model(unet2d_nuclei_broad_model):
     assert test_model(unet2d_nuclei_broad_model)
 
 
-@pytest.mark.skipif(pytest.skip_torch, reason="requires torch")
-def test_predict_image(unet2d_nuclei_broad_model, tmp_path):
+def test_predict_image(unet2d_fixed_shape_or_not, tmpdir):
```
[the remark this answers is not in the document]
constantinpape: I guess the model list is empty otherwise?

tests/weight_converter/torch/test_torchscript.py
```diff
@@ -3,20 +3,10 @@
 
 
 @pytest.mark.skipif(pytest.skip_torch, reason="requires pytorch")
```
constantinpape, line 5: Do we still need skips?

tests/conftest.py
```diff
@@ -2,75 +2,91 @@
...
+    ),
     "unet2d_multi_tensor": (
         "https://raw.githubusercontent.com/bioimage-io/spec-bioimage-io/main/example_specs/models/"
         "unet2d_multi_tensor/rdf.yaml"
     ),
-    "FruNet_model": "https://sandbox.zenodo.org/record/894498/files/rdf.yaml",
-    # "FruNet_model": "https://raw.githubusercontent.com/deepimagej/models/master/fru-net_sev_segmentation/model.yaml",
 }
 
+try:
```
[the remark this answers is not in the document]
FynnBe: to have `load_packages` available for deriving a fixture from...  the derived fixture is `any_model` which runs the test with all models that are not skipped due to missing imports.

tests/test_prediction.py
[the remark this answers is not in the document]
FynnBe: ah, no that is a bug! I was going to substitute it with `any_model`, but because the multi models don't work there yet that didn't fly...

tests/weight_converter/torch/test_torchscript.py
FynnBe: actually wherever specific test model groups are used the skip is still neccecarry as those groups are not shrinking to length one and will fail to provide the model (as it is not loaded).

maybe we can make those then empty lists instead, if the tests then pass/skip (haven't tried that yet)

constantinpape: Ok, I don't understand the new logic in the tests yet, could you quickly summarise it, or even better write a comment in conftest.py to explain it?

FynnBe: I'll add a comment...

FynnBe: just tested the empty list in a parametrized fixture and it works, meaning it just skips the test (and marks it skipped as well), so I'll use that instead of the skip flags where possible

tests/conftest.py
FynnBe: now only `skip_onnx` is still in use due to some tests dependency on torch and onnx...

FynnBe:
> * `FAILED tests/weight_converter/torch/test_onnx.py::test_onnx_converter_from_torch[unet2d_multi_tensor]` I don't understand why this one fails, it passes on main!

on main it is only tested with the previous unet2d_nuclei_broad_model, here the test fails for the unet2d_multi_tensor, which wasn't tested before...

constantinpape:
> on main it is only tested with the previous unet2d_nuclei_broad_model, here the test fails for the unet2d_multi_tensor, which wasn't tested before...

But I have added this test before: https://github.com/bioimage-io/python-bioimage-io/blob/d41dc9955ee6211e0ac854c500050b98e650f1f4/tests/weight_converter/torch/test_onnx.py#L16-L24

Which  used the multi tensor model and did pass!

FynnBe:
> Which used the multi tensor model and did pass!

yep, no idea how and why that wasn't the case, but now it passes here again, too.

The conda-build job fails, but has not run on main for the past few months, so I'm merging

PR #106, opened 2021-09-15
Simplify PredictionPipeline

FynnBe:
builds ontop of #105

### DIFF
--- a/bioimageio/core/prediction_pipeline/_prediction_pipeline.py
+++ b/bioimageio/core/prediction_pipeline/_prediction_pipeline.py
@@ -257,7 +257,7 @@ def create_prediction_pipeline(
         elif isinstance(out.shape, ImplicitOutputShape):
             named_output_shape.append(
                 NamedImplicitOutputShape(
-                    reference_input=out.shape.reference_input,
+                    reference_input=out.shape.reference_tensor,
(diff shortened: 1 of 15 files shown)


bioimageio/core/prediction_pipeline/_prediction_pipeline.py
```diff
@@ -144,26 +91,6 @@ def input_specs(self):
     def output_specs(self):
         return self._output_specs
 
-    @property
```
k-dominik, line 147: so, with these properties gone, are they supposed to be queried from the model?

k-dominik (approved):
cool to have all the NN dependency checking for tests in a single place now :)

bioimageio/core/prediction_pipeline/_prediction_pipeline.py
FynnBe: they are supposed to be accessed through `input_specs` and `output_specs` (which are set from `model.inputs` and `model.outputs` respectively. This avoids redefining everything as the (non-raw/derived) nodes are now part of this core module.

k-dominik: good to know. Sounds reasonable :)

PR #108, opened 2021-09-16
Processing with dataset statistics and reference_tensor

FynnBe:


constantinpape:
@FynnBe before implementing anything, can you maybe sketch out your proposed solution here?

My initial thoughts are:
- why do we need the preprocess dataclasses
- can we do this with less changes to the initial design?

When I thought about this, I had something like this in mind (very rough sketch!):
```python
# there is a `Statistics` class that gives access to the required statistics
# we could also use the `Measure` you propose instead

class Statistics
   @property
   def mean
   ...

# processing definitions all take a `dict[Statistics]`, mapping from tensor name to stats object as optional argument, e.g. 
def zero_mean_unit_variance(.., stats: Optional[dict[Statistics]]):
  if mode == "per_dataset":
      mean = stats[tensor_name].mean

# usage: there are helper functions to determine the necessary stats given all processing and a helper function to compute them (needs to be separate functions so that users can implement more efficient versions or cover cases where we don't understand what the "dataset" is)
# the resulting stats dict is passed to prediction pipeline (either in constructor or as argument for predict), which passes it on to the processing functions
```

What's still missing in that design: how do processing functions know which tensor they act upon (relevant for the ones without reference tensor).

Also I don't mind at all going with another solution, but I would like to discuss possible designs before implementation! Ideally we find something with minimal changes that would allow to release without implementing everything.

bioimageio/core/prediction_pipeline/statistical_measures.py
```diff
@@ -0,0 +1,30 @@
...
+
+@dataclass(frozen=True)
+class Std(Measure):
+    value: Optional[float] = None
+
+
+@dataclass(frozen=True)
+class Percentile(Measure):
+    n: float
+    value: Optional[float] = None
```
k-dominik, line 25: what would be the meaning of `value` here?

bioimageio/core/prediction_pipeline/_processing.py
```diff
@@ -0,0 +1,228 @@
...
+def ensure_dtype(tensor: xr.DataArray, *, dtype) -> xr.DataArray:
+    """
+    Convert array to a given datatype
+    """
+    return tensor.astype(dtype)
+
+
+@dataclass
+class Processing:
+    apply_to: str
```
k-dominik, line 37: not all preprocessings use the `apply_to` parameter.

bioimageio/core/prediction_pipeline/_processing.py
```diff
@@ -0,0 +1,228 @@
...
+
+@dataclass
+class ScaleLinear(Processing):
+    """scale the tensor with a fixed multiplicative and additive factor"""
+
+    gain: float
+    offset: float
+    axes: Optional[Sequence[str]] = None
+
+    def apply(self, tensor: xr.DataArray) -> xr.DataArray:
```
k-dominik, line 95: should probably be `tensors` and make use of `apply_to`?

FynnBe:
> why do we need the preprocess dataclasses

I know you prefer functions, but I think classes can help sharing common pattern across the processings. Dataclasses just further reduce the boiler plate code, imho

>What's still missing in that design: how do processing functions know which tensor they act upon (relevant for the ones without reference tensor).

that's where the `apply_to` comes in

bioimageio/core/prediction_pipeline/statistical_measures.py
FynnBe: the computed percentile

bioimageio/core/prediction_pipeline/_processing.py
FynnBe: they have to (should)

bioimageio/core/prediction_pipeline/_processing.py
FynnBe: yeah, or rather apply_simple...

bioimageio/core/prediction_pipeline/_processing.py
```diff
@@ -0,0 +1,228 @@
...
+
+    def apply_simple(self, tensor: xr.DataArray) -> xr.DataArray:
+        """apply processing to single tensor"""
+        raise NotImplementedError
+
+    def __post_init__(self):
+        """validate common kwargs by their annotations"""
+        if hasattr(self, "mode"):
+            if self.mode not in get_args(self.mode):
+                raise NotImplementedError(f"Unsupported mode {self.mode} for {self.__class__.__name__}: {self.mode}")
```
FynnBe, line 84: This is basically the whole sketch: with this base we can hide all the tensor name and stats complexity for simple cases and have convenient access to them when we need it...

bioimageio/core/prediction_pipeline/_processing.py
```diff
@@ -0,0 +1,228 @@
...
+
+        return tensors
+
+
+@dataclass
+class Binarize(Processing):
+    threshold: float
+
+    def apply_simple(self, tensor: xr.DataArray) -> xr.DataArray:
+        return ensure_dtype(tensor > self.threshold, dtype="float32")
```
FynnBe, line 149: simple case using `apply_simple`

bioimageio/core/prediction_pipeline/_processing.py
```diff
@@ -0,0 +1,228 @@
...
+            if self.axes:
+                axes = tuple(self.axes)
+            else:
+                axes = None
+
+            v_lower = ref_tensor.quantile(self.min_percentile / 100.0, dim=axes)
+            v_upper = ref_tensor.quantile(self.max_percentile / 100.0, dim=axes)
+        elif self.mode == "per_dataset":
+            v_lower = self.get_computed_statistics(ref_name, "v_lower")
+            v_upper = self.get_computed_statistics(ref_name, "v_upper")
```
FynnBe, line 182: somewhat convenient access to stats

bioimageio/core/prediction_pipeline/_processing.py
```diff
@@ -0,0 +1,228 @@
...
+    reference_tensor: Optional[str] = None
+
+    def get_required_statistics(self) -> Dict[str, Dict[str, Measure]]:
+        if self.mode == "per_sample":
+            return {}
+        elif self.mode == "per_dataset":
+            measures = {"v_lower": Percentile(self.min_percentile), "v_upper": Percentile(self.max_percentile)}
+            return {self.reference_tensor or self.apply_to: measures}
+        else:
+            raise ValueError(self.mode)
```
FynnBe, line 167: defining what stats exactly are needed

bioimageio/core/prediction_pipeline/_processing.py
```diff
@@ -0,0 +1,228 @@
...
+            raise RuntimeError(f"Missing {measure_name} for {tensor_name} dataset.")
+
+        return ret
+
+    def apply(self, **tensors: xr.DataArray) -> Dict[str, xr.DataArray]:
+        """apply processing to named tensors; call 'apply_simple' as default"""
+        tensors[self.apply_to] = self.apply_simple(tensors[self.apply_to])
+        return tensors
+
+    def apply_simple(self, tensor: xr.DataArray) -> xr.DataArray:
```
k-dominik, line 76: make private

FynnBe:
notes from meeting on this:
- remove `.value` from `Measure` (use dict)
- remove measure names from required/computed statistics dict
for later:
- add `copy_tenosr` to 'save' referenceable tensors along the processing

constantinpape:
One more thought on this: to compute and identify statistics, we need some of the processing kwargs in addition to the measure itself. Because this can be different depending on `axes` (not sure if other kwargs are relevant, probably not).

FynnBe:
> One more thought on this: to compute and identify statistics, we need some of the processing kwargs in addition to the measure itself. Because this can be different depending on `axes` (not sure if other kwargs are relevant, probably not).

Not sure if I fully understand your point, but isn't this taken care of by the fact that each initialized, derived Processing specifies the required statistics by overwriting the `get_required_dataset_statistics` (and it's upcoming sample version) method....

constantinpape:
> Not sure if I fully understand your point, but isn't this taken care of by the fact that each initialized, derived Processing specifies the required statistics by overwriting the `get_required_dataset_statistics` (and it's upcoming sample version) method.?

But we might have one processing that requires `mean(tensor1, axes=None)` and the other that requires `mean(tensor1, axes="xy")`. So these are two mean requests that are different (and in fact one is a scalar and the other a list).

As far as I understood what we discussed this morning, we wanted to identify statistics with a `tensor-name` (as dict key) and `Measure` (also as dict key). And in this case, if we just have `Mean`, we cannot distinguish between different axes.
(We could of course add `axes` as member to `Mean(Measure)`, as we said that we use the instantiated classes as keys).

Hope that makes my point clear.

FynnBe:
> But we might have one processing that requires `mean(tensor1, axes=None)` and the other that requires `mean(tensor1, axes="xy")`. So these are two mean requests that are different (and in fact one is a scalar and the other a list).
> 
> As far as I understood what we discussed this morning, we wanted to identify statistics with a `tensor-name` (as dict key) and `Measure` (also as dict key). And in this case, if we just have `Mean`, we cannot distinguish between different axes.
> (We could of course add `axes` as member to `Mean(Measure)`, as we said that we use the instantiated classes as keys).
> 
> Hope that makes my point clear.


oh right, of course! we should give `Mean` an `axes` parameter just like the `n` for `Percentile`

constantinpape:
> oh right, of course! we should give `Mean` an `axes` parameter just like the `n` for `Percentile`

Yes, that's def. the right solution.

FynnBe:
how about `tensor_name` instead of `apply_to` as attribute to `Processing`?

bioimageio/core/prediction_pipeline/_combined_processing.py
```diff
@@ -0,0 +1,183 @@
...
+    def apply_postprocessing(
+        self, *output_tensors: xr.DataArray, input_sample_statistics: Dict[str, Dict[Measure, Any]]
+    ) -> Tuple[List[xr.DataArray], Dict[str, Dict[Measure, Any]]]:
+        assert len(output_tensors) == len(self.output_tensor_names)
+        tensors = dict(zip(self.output_tensor_names, output_tensors))
+        sample_stats = input_sample_statistics
+        sample_stats.update(self.compute_sample_statistics(tensors, self._req_output_stats[SAMPLE]))
+        for proc in self._post:
+            proc.set_computed_sample_statistics(sample_stats)
+            tensors[proc.tensor_name] = proc.apply(tensors[proc.tensor_name])
```
constantinpape, line 116: I don't see where we make sure that the postprocessed tensors have the correct dtype.
(After postprocessing we need to return the same dtype as is specified in the output tensor spec.)

tests/prediction_pipeline/test_postprocessing.py
```diff
@@ -13,8 +13,6 @@ def test_binarize_postprocessing():
     threshold = 0.5
     exp = xr.DataArray(np_data > threshold, dims=axes)
 
-    for dtype in ("float32", "float64", "uint8", "uint16"):
```
constantinpape, line 16: Don't remove the test for different dtypes here. We still need to ensure that postprocessing yields the correct dtype that is specified in the output tensor spec.

constantinpape (commented):
As far as I can see this is not handling the dtype in the output tensor spec yet.
Looks good otherwise.

bioimageio/core/prediction_pipeline/_combined_processing.py
FynnBe: here: https://github.com/bioimage-io/core-bioimage-io-python/pull/108/files#diff-1d7498945a046a703ef8456b221372224ab34bc5251067c2ba0bc7bf13025e5fR64-R69

tests/prediction_pipeline/test_postprocessing.py
FynnBe: the helpers `make_pre/postprocessing` do not well into the new code, thus `CombinedProcessing` now appends an `EnsureDtype` processing at the end. (Maybe we want to add that one to the spec as well so it can be used between processing steps as well?
I'll add a test for the output dtype using `CombinedPreprocessing`

FynnBe:
no idea how you mentioned to comment on this: https://github.com/bioimage-io/core-bioimage-io-python/pull/108#discussion_r710882772
but I have renamed it to `tensor_name` in the meantime...

bioimageio/core/prediction_pipeline/_processing.py
FynnBe: ah, here.. `apply_to` has been, renamed to `tensor_name`

tests/prediction_pipeline/test_postprocessing.py
constantinpape: > Maybe we want to add that one to the spec as well so it can be used between processing steps as well?

I don't know what you mean by this, but I would propose to do spec changes later and focus on implementing the current spec in this PR ;).

constantinpape (approved):
LGTM now.

tests/prediction_pipeline/test_postprocessing.py
FynnBe: yes of course. I mean that we might want to think about adding another processing to the documented bioimageio.spec: `EnsureDtype` or better `CastDtype`. Here we use the (secret) `EnsureDtype` Processing to ensure that at the end of postprocessing the tensor is cast to the specified dtype. 
What is new is that we do not cast in-between postprocessing steps anymore (and I don't think that is required).

constantinpape: > `EnsureDtype` or better `CastDtype`. Here we use the (secret) `EnsureDtype` Processing to ensure that at the end of postprocessing the tensor is cast to the specified dtype.

Yes, we can think about adding this to the spec at some point.

> What is new is that we do not cast in-between postprocessing steps anymore (and I don't think that is required).

As far as I can recall, we didn't cast in between in the previous implementation either and would just cast at the end.

PR #111, opened 2021-09-20
run tests via tox before publishing to pypi

k-dominik:
added tox config for testing in a non-conda environment.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,11 +26,17 @@ jobs:
           python -m pip install --upgrade pip
           pip --version
 
-      - name: Install dependencies
+      - name: install tox
+        run: |
+          pip install tox
+
+      - name: run tests
+        run: |
+          tox
+
+      - name: Install build dependencies
         run: |
           pip install wheel
-  
-      # TODO: add tests here
 
       - name: Check if there is a parent commit
         id: check-parent-commit
--- a/setup.py
+++ b/setup.py
@@ -25,10 +25,10 @@ setup(
         "Programming Language :: Python :: 3.8",
     ],
     packages=find_namespace_packages(exclude=["tests"]),  # Required
-    install_requires=["bioimageio.spec", "imageio>=2.5", "numpy", "xarray"],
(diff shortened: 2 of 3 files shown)


oeway:
Hi @k-dominik Thanks for looking into this. Did you run pytest locally? 

It would be nice to still have a requirements.txt file with all the versions pined for testing with CI.

k-dominik:
> Hi @k-dominik Thanks for looking into this. Did you run pytest locally?
> 
> It would be nice to still have a requirements.txt file with all the versions pined for testing with CI.

sure, I ran them locally. Why the pinning, Wei? I argue a single source of truth for deps. If we now start adding _another_ file, with different dependencies.... Also what's the purpose? Tox will use setup.py...

oeway:
> > Hi @k-dominik Thanks for looking into this. Did you run pytest locally?
> > It would be nice to still have a requirements.txt file with all the versions pined for testing with CI.
> 
> sure, I ran them locally. Why the pinning, Wei? I argue a single source of truth for deps. If we now start adding _another_ file, with different dependencies.... Also what's the purpose? Tox will use setup.py...

@MartinHjelmare has more experiences in best practices for python projects, and recommended that to us for the ImJoy python projects (see [our template](https://github.com/imjoy-team/imjoy-python-template)). As I understand, pining all version in the requirements for CI testing is a recommended because it guarantees a stable and reproducible testing environment. E.g. we will save time in debugging if we have the exact test environment in the CI and locally, today and tomorrow etc. In general, we should not pin version in the setup.py, but for testing it's good to pin the versions.

k-dominik:
I think that's overkill - especially with all those additional dependencies (that we don't even test via pip). It adds to maintenance burden and also to developers working on different environments than users see in production. I'd agree to test against something like this with a clear deployment target.

oeway:
I will merge this for now!

We can continue discussion about whether we should pin, or delay it when we see issues related to that.

k-dominik:
yeah let's discuss this in the next meeting.

PR #115, opened 2021-09-27
accept URLs in test-model command

FynnBe:
previously input was always cast to pathlib.Path which does not work for urls.


k-dominik (approved):
awesome, - so that works already for package?! (would be awesome)...

FynnBe:
yes, for package it works already

PR #117, opened 2021-09-28
Update build_model for changed output shape spec; add test

constantinpape:


constantinpape:
This is just a minor fix + test coverage increase, so I am merging.

PR #118, opened 2021-09-28
Update VERSION

constantinpape:
I suggest to bump the version to include some fixes since the first release.
After merging this we need to check that CI does a correct job and also creates a GH release (did not happen correctly the first time).


oeway (approved):

oeway:
Let's try it!

oeway:
It works!
https://github.com/bioimage-io/core-bioimage-io-python/releases/tag/v0.4.1

PR #121, opened 2021-10-07
Doi support

FynnBe:
requires https://github.com/bioimage-io/spec-bioimage-io/pull/251

accept dois as rdf sources. 

For now this will only work for zenodo sandbox dois. It is very straight-forward to add support for regular dois that point to zenodo (we just need a non-sandbox test doi for a test case and some minor code adjustments), but outside of zenodo it is unclear how to resolve the url to the actual `rdf.yaml` (or zipped package) in case the doi points to a more absttract entry (like for zenodo the record with all files, including the `rdf.yaml`. 

Zenodo has an API to find the url to the `rdf.yaml` (atm we don't expect zipped packages to be hosted on zenodo), but for "any" doi it is unclear what to do. The default is now assuming that the doi directly resolves to an `rdf.yaml` or a zipped package.


FynnBe:
> @FynnBe this is still failing in the tests versus main.

I don't really get why the import of def `resolve_rdf_source` fails. It's right there:
https://github.com/bioimage-io/spec-bioimage-io/blob/main/bioimageio/spec/io_.py#L35

PR #130, opened 2021-10-19
Add test_resource

FynnBe:
an extended version of `test_model`.
No checks are performed for non-model resources yet (but they won't raise an exception)

Todo:
 - [ ] decide wether `test_model` shoulda also return a summary dict instead of a boolean...

### DIFF
--- a/bioimageio/core/prediction.py
+++ b/bioimageio/core/prediction.py
@@ -4,22 +4,29 @@ import warnings
 from copy import deepcopy
 from itertools import product
 from pathlib import Path
-from typing import Dict, List, OrderedDict, Sequence, Tuple, Union
+from typing import Dict, List, Optional, OrderedDict, Sequence, Tuple, Union
 
 import imageio
 import numpy as np
 import xarray as xr
-
-from bioimageio.core import load_resource_description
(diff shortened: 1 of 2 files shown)


bioimageio/core/prediction.py
```diff
@@ -476,30 +484,71 @@ def predict_images(
...
+    weight_format: Optional[WeightsFormat] = None,
+    devices: Optional[List[str]] = None,
+    decimal: int = 4,
+) -> bool:
     """Test whether the test output(s) of a model can be reproduced.
 
     Returns True if the test passes, otherwise returns False and issues a warning.
     """
     model = load_resource_description(model_rdf)
     assert isinstance(model, Model)
```
constantinpape, line 498: This should probably not be an assertion but a more meaningfull error.

constantinpape (commented):
I think `test_model` should also return a dict. (Then we slightly need to adapt the CLI code).
Also, `test_resource` should also get a CLI function, and we may consider refactoring `test_model` and `test_resource` to a separate file; `prediction.py` seems not quite fitting any more if we test for different resources as well.

PR #134, opened 2021-10-20
also accept model.yaml in place of rdf.yaml

FynnBe:
needs https://github.com/bioimage-io/spec-bioimage-io/pull/261


FynnBe:
or do we pin on build automatically?

PR #136, opened 2021-10-30
Throw exception if `run_mode` is specified

FynnBe:
as we have no `run_mode` implemented we should throw a `NotImpllementedError` in that case...


constantinpape (approved):
In the future we may want to support some known run modes in the validation as well, but for now this is good.

PR #138, opened 2021-11-03
Support more dependency managers in build_model

constantinpape:


constantinpape:
This is just a minor improvement and some bugfixes in `build_model`, so I am gonna merge it now.

PR #141, opened 2021-11-11
Add debug_model function

constantinpape:
This adds a helper function to debug a model, which returns test inputs, outputs, expected outputs as well as relevant intermediates.
In order to implement it I have changed the pre-and-postprocessing calls in `PredictionPipeline` s.t. they also return the stats parameter; otherwise it's not possible to use the postprocessing function.


FynnBe:
returning the stats is good and it's just internal anyway. 👍

PR #142, opened 2021-11-11
Update scale range

constantinpape:
Our implementation of `scale_range` is different than the one in csbdeep: https://github.com/CSBDeep/CSBDeep/blob/master/csbdeep/utils/utils.py#L51-L75.
Hence we cannot reproduce the stardist models right now, see also https://github.com/stardist/stardist/pull/171.

@FynnBe I am not sure what's the best strategy here: we can either just change the function or add a flag for the CSBDeep style normalization.

PR builds on #141.

### DIFF
--- a/bioimageio/core/prediction_pipeline/_processing.py
+++ b/bioimageio/core/prediction_pipeline/_processing.py
@@ -166,6 +166,7 @@ class ScaleRange(Processing):
     axes: Optional[Sequence[str]] = None
     min_percentile: float = 0.0
     max_percentile: float = 100.0
+    eps: float = 1e-6
     reference_tensor: Optional[str] = None
 
     def get_required_dataset_statistics(self) -> Dict[str, Set[Measure]]:
@@ -209,7 +210,7 @@ class ScaleRange(Processing):
         v_lower = get_stat(ref_name, Percentile(self.min_percentile, axes=axes))
         v_upper = get_stat(ref_name, Percentile(self.max_percentile, axes=axes))
 
-        return ensure_dtype((tensor - v_lower) / (v_upper - v_lower), dtype="float32")
+        return ensure_dtype((tensor - v_lower) / (v_upper - v_lower + self.eps), dtype="float32")
 
     def __post_init__(self):
         super().__post_init__()
--- a/bioimageio/core/prediction_pipeline/_processing.py
+++ b/bioimageio/core/prediction_pipeline/_processing.py
@@ -209,7 +209,7 @@ class ScaleRange(Processing):
         v_lower = get_stat(ref_name, Percentile(self.min_percentile, axes=axes))
         v_upper = get_stat(ref_name, Percentile(self.max_percentile, axes=axes))
 
-        return ensure_dtype((tensor - v_lower) / v_upper, dtype="float32")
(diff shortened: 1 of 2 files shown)


uschmidt83:
I couldn't find the actual code where you compute the percentile values of the tensor / data array.

bioimageio/core/prediction_pipeline/_processing.py
```diff
@@ -209,7 +209,7 @@ def apply(self, tensor: xr.DataArray) -> xr.DataArray:
         v_lower = get_stat(ref_name, Percentile(self.min_percentile, axes=axes))
         v_upper = get_stat(ref_name, Percentile(self.max_percentile, axes=axes))
 
-        return ensure_dtype((tensor - v_lower) / v_upper, dtype="float32")
+        return ensure_dtype((tensor - v_lower) / (v_upper - v_lower), dtype="float32")
```
FynnBe, line 212: IMO this update makes a lot of sense to actually map the max percentile to 1...
I'm inclined to view the current implementation as buggy and would like to just fix/update it.
The spec is not specific: https://github.com/bioimage-io/spec-bioimage-io/blob/main/supported_formats_and_operations.md#preprocessing
 @esgomezm @tomburke-rse what are your implementations/expectations. Can we just change it?

constantinpape:
@uschmidt83 the percentiles are computed here: https://github.com/bioimage-io/core-bioimage-io-python/blob/main/bioimageio/core/prediction_pipeline/_combined_processing.py#L165-L166

This is done not directly in the normalization function, because we enable scaling a tensor based on the percentiles computed from another tensor (we had some use-cases that require this).

bioimageio/core/prediction_pipeline/_processing.py
tomburke-rse: In imagej-modelzoo, the percentile normalization already does this as far as I can see:
https://github.com/imagej/imagej-modelzoo/blob/master/src/main/java/net/imagej/modelzoo/plugin/transformation/preprocessing/NormalizePercentilePreprocessing.java

esgomezm: We also divide using v_lower: https://github.com/deepimagej/imagej-macros/blob/648caa867f6ccb459649d4d3799efa1e2e0c5204/bioimage.io/per_sample_scale_range.ijm#L92

```
	//normalization
	diff = ma-mi+1e-20; // add epsilon to avoid 0-divisions
	run("32-bit");
	run("Subtract...", "value="+mi+" stack");
	run("Divide...", "value="+diff+" stack");
```

constantinpape: Ok, then let's update it!

FynnBe:
I suppose we should add an epsilon to scale range, too then...
I'll add it in the spec

constantinpape:
> I suppose we should add an epsilon to scale range, too then...
> I'll add it in the spec

Ok, I will wait for this and integrate it here then as well.

constantinpape:
@FynnBe I think this is good to go; anything to add from your side?

FynnBe (approved):
LGTM 👍

PR #146, opened 2021-11-12
make some Arguments Options

FynnBe:
We use a lot of `typer.Argument` in the CLI. I think some of the input should rather be a `typer.Option`.
E.g. for `test-model` was:
```
$ bioimageio test-model --help
bioimageio.spec 0.3.4
implementing:
        general RDF 0.2.0
        model RDF 0.3.4
bioimageio.core 0.4.3
Usage: __main__.py test-model [OPTIONS] MODEL_RDF [WEIGHT_FORMAT]:[pytorch_sta
                              te_dict|pytorch_script|keras_hdf5|tensorflow_js|
                              tensorflow_saved_model_bundle|onnx] [DEVICES]...
                              [DECIMAL]

  Test whether the test output(s) of a model can be reproduced.

  Returns summary dict with "error" and "traceback" key; summary["error"] is
  None if no errors were encountered.

Arguments:
  MODEL_RDF                       Path or URL to the model resource
                                  description file (rdf.yaml) or zipped model.
                                  [required]

  [WEIGHT_FORMAT]:[pytorch_state_dict|pytorch_script|keras_hdf5|tensorflow_js|tensorflow_saved_model_bundle|onnx]
                                  The weight format to use.
  [DEVICES]...                    Devices for running the model.
  [DECIMAL]                       The test precision.  [default: 4]

Options:
  -h, --help, --version  Show this message and exit.

```
and is now:
```
$ bioimageio test-model --help
bioimageio.spec 0.3.4
implementing:
        general RDF 0.2.0
        model RDF 0.3.4
bioimageio.core 0.4.3
Usage: __main__.py test-model [OPTIONS] MODEL_RDF

  Test whether the test output(s) of a model can be reproduced.

  Returns summary dict with "error" and "traceback" key; summary["error"] is
  None if no errors were encountered.

Arguments:
  MODEL_RDF  Path or URL to the model resource description file (rdf.yaml) or
             zipped model.  [required]


Options:
  --weight-format [pytorch_state_dict|pytorch_script|keras_hdf5|tensorflow_js|tensorflow_saved_model_bundle|onnx]
                                  The weight format to use.
  --devices TEXT                  Devices for running the model.
  --decimal INTEGER               The test precision.  [default: 4]
  -h, --help, --version           Show this message and exit.

```

I think this makes it easier to read and use.

Issue #144, opened 2021-11-11
Can't select weight format in test_model (and other CLI?)


constantinpape:
It looks like the new weight format selection mechanism is broken:
```
$ bioimageio test-model model.yaml tensorflow_saved_model_bundle
bioimageio.spec 0.3.4
implementing:
        general RDF 0.2.0
        model RDF 0.3.3
bioimageio.core 0.4.3post0
/home/pape/Work/bioimageio/spec-bioimage-io/bioimageio/spec/rdf/v0_2/schema.py:194: UserWarning: BSD-2 is not a recognized SPDX license identifier. See https://spdx.org/licenses/
  warnings.warn(f"{value} is not a recognized SPDX license identifier. See https://spdx.org/licenses/")
Model test for model.yaml has FAILED!
{'error': "Weight format WeightFormatEnum.tensorflow_saved_model_bundle is not in supported formats ['pytorch_state_dict', 'tensorflow_saved_model_bundle', 'pytorch_script', 'onnx', 'keras_hdf5']", 'traceback': ['  File "/home/pape/Work/bioimageio/python-bioimage-io/bioimageio/core/resource_tests.py", line 56, in test_resource\n    bioimageio_model=model, devices=devices, weight_format=weight_format\n', '  File "/home/pape/Work/bioimageio/python-bioimage-io/bioimageio/core/prediction_pipeline/_prediction_pipeline.py", line 149, in create_prediction_pipeline\n    bioimageio_model=bioimageio_model, devices=devices, weight_format=weight_format\n', '  File "/home/pape/Work/bioimageio/python-bioimage-io/bioimageio/core/prediction_pipeline/_model_adapters/_model_adapter.py", line 53, in create_model_adapter\n    raise ValueError(f"Weight format {weight_format} is not in supported formats {_WEIGHT_FORMATS}")\n']}
```
@FynnBe please also add a test to the CLI tests that tests this so we don't run into this issue again without noticing!

PR #147, opened 2021-11-12
Fix specifying weight format in CLI (#144)

FynnBe:
and add more CLI tests

needs small update if #146 is merged (argument -> option)

### DIFF
--- a/bioimageio/core/__main__.py
+++ b/bioimageio/core/__main__.py
@@ -77,7 +77,12 @@ def test_model(
     # this is a weird typer bug: default devices are empty tuple although they should be None
     if len(devices) == 0:
         devices = None
-    summary = resource_tests.test_model(model_rdf, weight_format=weight_format.value, devices=devices, decimal=decimal)
+    summary = resource_tests.test_model(
+        model_rdf,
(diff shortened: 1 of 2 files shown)


bioimageio/core/__main__.py
```diff
@@ -102,7 +102,7 @@ def test_resource(
     # this is a weird typer bug: default devices are empty tuple although they should be None
     if len(devices) == 0:
         devices = None
-    summary = resource_tests.test_resource(rdf, weight_format=weight_format, devices=devices, decimal=decimal)
+    summary = resource_tests.test_resource(rdf, weight_format=weight_format.value, devices=devices, decimal=decimal)
```
constantinpape, line 105: This does not work if `weight_format` is `None`. (That's also why the tests currently fail).

constantinpape (commented):
Tests still fail because `weight_format.value` is invalid for the default value (=`None`).
Note that we need the same fixes for all CLI functions that take weight_format as an argument (so also `predict_image` and `predict_images`.)

constantinpape (approved):
LGTM from my side; feel free to merge once you think this is ready @FynnBe

PR #151, opened 2021-11-20
Add deepimagej config generation

constantinpape:
Write the deepimagej config. Still need to get the pixel size information somehow.
cc @esgomezm

### DIFF
--- a/bioimageio/core/build_spec/build_model.py
+++ b/bioimageio/core/build_spec/build_model.py
@@ -1,7 +1,6 @@
 import datetime
 import hashlib
 import os
-import shutil
 from pathlib import Path
 from shutil import copyfile
 from typing import Any, Dict, List, Optional, Tuple, Union
@@ -291,7 +290,8 @@ def _get_deepimagej_macro(name, kwargs, export_folder):
     url = f"https://raw.githubusercontent.com/deepimagej/imagej-macros/master/bioimage.io/{macro}"
 
     path = os.path.join(export_folder, macro)
-    # TODO do we use requests?
+    # use https://github.com/bioimage-io/core-bioimage-io-python/blob/main/bioimageio/core/resource_io/utils.py#L267
+    # instead if the implementation is update s.t. an output path is accepted
     with requests.get(url, stream=True) as r:
         with open(path, "w") as f:
             f.write(r.text)
--- a/bioimageio/core/build_spec/build_model.py
+++ b/bioimageio/core/build_spec/build_model.py
@@ -1,6 +1,7 @@
 import datetime
 import hashlib
 import os
+import shutil
 from pathlib import Path
 from shutil import copyfile
 from typing import Any, Dict, List, Optional, Tuple, Union
@@ -11,6 +12,7 @@ import requests
 import bioimageio.spec as spec
 import bioimageio.spec.model as model_spec
 from bioimageio.core import export_resource_package, load_raw_resource_description
+from bioimageio.core.resource_io.nodes import URI
 from bioimageio.core.resource_io.utils import resolve_local_source, resolve_source
 
 try:
@@ -52,8 +54,8 @@ def _infer_weight_type(path):
         raise ValueError(f"Could not infer weight type from extension {ext} for weight file {path}")
 
 
-def _get_weights(weight_uri, weight_type, source, root, **kwargs):
-    weight_path = resolve_source(weight_uri, root)
+def _get_weights(original_weight_source, weight_type, source, root, **kwargs):
+    weight_path = resolve_source(original_weight_source, root)
     if weight_type is None:
         weight_type = _infer_weight_type(weight_path)
     weight_hash = _get_hash(weight_path)
@@ -65,7 +67,7 @@ def _get_weights(weight_uri, weight_type, source, root, **kwargs):
         source_file, source_class = source.replace("::", ":").split(":")
 
         # get the source path
-        source_file = resolve_source(source_file, root)
+        source_file = _ensure_local(source_file, root)
         source_hash = _get_hash(source_file)
 
         # if not relative, create local copy (otherwise this will not work)
@@ -85,7 +87,9 @@ def _get_weights(weight_uri, weight_type, source, root, **kwargs):
         attachments = {}
 
     weight_types = model_spec.raw_nodes.WeightsFormat
-    weight_source = resolve_local_source(weight_uri, root)
+    weight_source = _ensure_local_or_url(original_weight_source, root)
+
+    assert isinstance(weight_source, URI) or weight_source.is_relative_to(root)
     if weight_type == "pytorch_state_dict":
         # pytorch-state-dict -> we need a source
         assert source is not None
@@ -135,7 +139,7 @@ def _get_weights(weight_uri, weight_type, source, root, **kwargs):
 
     elif weight_type == "tensorflow_js":
         weights = model_spec.raw_nodes.TensorflowJsWeightsEntry(
-            source=weight_uri,
+            source=weight_source,
             sha256=weight_hash,
             tensorflow_version=kwargs.get("tensorflow_version", "1.15"),
             **attachments,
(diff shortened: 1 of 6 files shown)


esgomezm:
Hi @constantinpape 

In the ZeroCostDL4Mic notebook for StarDist we directly ask the user for this value. By default the pixel size could be set to 1 pixel. 

In the rest of the notebooks, it's calculated using the following function:
```python
from PIL import Image 
def getPixelSizeTIFFmetadata(TIFFpath, display=False):
  with Image.open(TIFFpath) as img:
    meta_dict = {TAGS[key] : img.tag[key] for key in img.tag.keys()}


  # TIFF tags
  # https://www.loc.gov/preservation/digital/formats/content/tiff_tags.shtml
  # https://www.awaresystems.be/imaging/tiff/tifftags/resolutionunit.html
  ResolutionUnit = meta_dict['ResolutionUnit'][0] # unit of resolution
  width = meta_dict['ImageWidth'][0]
  height = meta_dict['ImageLength'][0]

  xResolution = meta_dict['XResolution'][0] # number of pixels / ResolutionUnit

  if len(xResolution) == 1:
    xResolution = xResolution[0]
  elif len(xResolution) == 2:
    xResolution = xResolution[0]/xResolution[1]
  else:
    print('Image resolution not defined.')
    xResolution = 1

  if ResolutionUnit == 2:
    # Units given are in inches
    pixel_size = 0.025*1e9/xResolution
  elif ResolutionUnit == 3:
    # Units given are in cm
    pixel_size = 0.01*1e9/xResolution
  else: 
    # ResolutionUnit is therefore 1
    print('Resolution unit not defined. Assuming: um')
    pixel_size = 1e3/xResolution

  if display:
    print('Pixel size obtained from metadata: '+str(pixel_size)+' nm')
    print('Image size: '+str(width)+'x'+str(height))
  
  return (pixel_size, width, height)
```

constantinpape:
Thanks for sharing the script @esgomezm, but I fear that this is not very useful for us because we can't rely on having tif files as inputs; the (mandatory) input data are numpy arrays here.
I will go ahead and just add the pixel size as a parameter to the `build_model` function and propagate it to the deepimagej config; then you can pass it to the function.

FynnBe:
incorporating pixel size into the spec (0.4.?) sounds like another hackathon item to me...

constantinpape:
> incorporating pixel size into the spec (0.4.?) sounds like another hackathon item to me...

Yes, def. a point to discuss

constantinpape:
Ok, I updated the deepimagej config function to include the pixel size and fix several issues.
There are still a couple open points:
- I don't understand the deepimagej binarize macro and how to set the threshold in it @esgomezm
- Does deepimagej support models with multiple inputs / outptus? If yes, how to specify the pre-post-processing for this in the config? Afaik there is just a single list of pre/postprocessing, which implies that this is applied to a single in/out tensors

Also, the tests are still failing due to some issues with sample inputs / outputs that I don't understand. But this may be fixed by #164, so we should merge that PR first and then merge main here.

FynnBe:
> Also, the tests are still failing due to some issues with sample inputs / outputs that I don't understand. But this may be fixed by #164, so we should merge that PR first and then merge main here.

yes, this should be fixed by #164

bioimageio/core/build_spec/build_model.py
```diff
@@ -258,6 +266,160 @@ def _get_dependencies(dependencies, root):
...
+
+    elif name == "zero_mean_unit_variance":
+        mode = kwargs["mode"]
+        if mode == "fixed":
+            macro = "fixed_zero_mean_unit_variance.ijm"
+            replace = {"paramMean": kwargs["mean"], "paramStd": kwargs["std"]}
+        else:
+            macro = "zero_mean_unit_variance.ijm"
+            replace = {}
+
```
FynnBe, line 281:
```suggestion
    else:
        raise ValueError(name)

```
would save you the duplication of available macro names in the assert above

bioimageio/core/build_spec/build_model.py
```diff
@@ -258,6 +266,160 @@ def _get_dependencies(dependencies, root):
...
+    # elif name == "binarize":
+    #     macro = "binarize.ijm"
+    #     replace = {"", kwargs["threshold"]}
+
+    macro = f"{name}.ijm"
+    url = f"https://raw.githubusercontent.com/deepimagej/imagej-macros/master/bioimage.io/{macro}"
+
+    path = os.path.join(export_folder, macro)
+    # TODO do we use requests?
+    with requests.get(url, stream=True) as r:
```
FynnBe, line 291: let me update https://github.com/bioimage-io/core-bioimage-io-python/blob/74e0f4beafabdb5a43df0d082dd3731c7f968441/bioimageio/core/resource_io/utils.py#L267 to accept `local_path` to get a nice tqdm pbar and then use it here...

constantinpape:
@FynnBe the tests for writing the deepimagej config is still failing with
```
 marshmallow.exceptions.ValidationError: {'sample_outputs': {0: ['Errors in all options for this field. Fix any of the following errors:', ['/tmp/bioimageio_cache/extracted_packages/8cfc68d44ac8c3a154290017d4151113277ca8a27e72a15b39fb4a0792b0ec79/sample_output_0.
tif'], ['expected relative path.']]}, 'sample_inputs': {0: ['Errors in all options for this field. Fix any of the following errors:', ['/tmp/bioimageio_cache/extracted_packages/8cfc68d44ac8c3a154290017d4151113277ca8a27e72a15b39fb4a0792b0ec79/sample_input_0.tif'], ['expected
 relative path.']]}} 
```
although sample_inputs/outputs are treated exactly the same as test_inputs/outputs in `build_model`.

bioimageio/core/build_spec/build_model.py
```diff
@@ -477,6 +705,17 @@ def build_model(
...
+    # debugging: for some reason this causes a validation error for the sample_input / output,
+    # although they are passed in the same format as test inputs / outputs:
+    # (Pdb) model.sample_inputs
+    # [PosixPath('/tmp/bioimageio_cache/extracted_packages/efae734f36ff8634977c9174f01c854c4e2cfc799bc92751a4558a0edd963da6/sample_input_0.tif')]
+    # (Pdb) model.sample_outputs
+    # [PosixPath('/tmp/bioimageio_cache/extracted_packages/efae734f36ff8634977c9174f01c854c4e2cfc799bc92751a4558a0edd963da6/sample_output_0.tif')]
+    # (Pdb) model.test_inputs
+    # [PosixPath('/tmp/bioimageio_cache/extracted_packages/efae734f36ff8634977c9174f01c854c4e2cfc799bc92751a4558a0edd963da6/test_input.npy')]
+    # (Pdb) model.test_outputs
+    # [PosixPath('/tmp/bioimageio_cache/extracted_packages/efae734f36ff8634977c9174f01c854c4e2cfc799bc92751a4558a0edd963da6/test_output.npy')]
```
FynnBe, line 718: should be fixed by #164 ; then this should be removed

FynnBe (commented):
almost there 👍

bioimageio/core/build_spec/build_model.py
FynnBe: > the tests for writing the deepimagej config is still failing with

... or not.. looking into it

FynnBe: the problem is that sample inputs/outputs are absolute and not relative paths...

constantinpape: but test inputs / outptus are also absolute paths. Why does it work for them and not for sample inputs/outputs?

FynnBe:
1942df0 contains bugs! not ready yet...

constantinpape:
> - I don't understand the deepimagej binarize macro and how to set the threshold in it @esgomezm

It's `optimal_thresold`.

> - Does deepimagej support models with multiple inputs / outptus? If yes, how to specify the pre-post-processing for this in the config? Afaik there is just a single list of pre/postprocessing, which implies that this is applied to a single in/out tensors

It's in theory possible, but too complex for now; we will just throw an error for now and leave this for future iterations.

bioimageio/core/build_spec/build_model.py
constantinpape: But the error message is more meaningful in the other case. I will update a bit differently.

FynnBe:
locally for me there remains an issue with sample_inputs/outputs serialization. Not sure what's going on there...

FynnBe:
> although sample_inputs/outputs are treated exactly the same as test_inputs/outputs in build_model.

this turned out to be wrong... in the deepimagej they are treated differently... (this should be fixed now though

FynnBe:
An update to bioimageio.spec should fix the remaining issue of serialization: https://github.com/bioimage-io/spec-bioimage-io/pull/294

constantinpape:
Thanks @FynnBe now it all works.

Issue #154, opened 2021-11-22
Tests fail on GPU (on windows)


FynnBe:
while the following test cases pass on cpu, they fail on gpu (on windows; only difference: `CUDA_VISIBLE_DEVICES=""`)
1. tests\prediction_pipeline\test_prediction_pipeline.py:36 (test_prediction_pipeline_torchscript[unet2d_multi_tensor])
2. tests\prediction_pipeline\test_prediction_pipeline.py:36 (test_prediction_pipeline_torchscript[unet2d_nuclei_broad_model])

@constantinpape I suspect this is independent of OS and only related to cuda+torchscript... 

1.:
```
tests\prediction_pipeline\test_prediction_pipeline.py:36 (test_prediction_pipeline_torchscript[unet2d_multi_tensor])
any_torchscript_model = WindowsPath('/Users/fbeut/bioimageio_cache/packages/multi-tensorp39.zip')

    def test_prediction_pipeline_torchscript(any_torchscript_model):
>       _test_prediction_pipeline(any_torchscript_model, "pytorch_script")

test_prediction_pipeline.py:38: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
test_prediction_pipeline.py:20: in _test_prediction_pipeline
    outputs = pp.forward(*inputs)
..\..\bioimageio\core\prediction_pipeline\_prediction_pipeline.py:129: in forward
    prediction = self.predict(*preprocessed)
..\..\bioimageio\core\prediction_pipeline\_prediction_pipeline.py:124: in predict
    return self._model.forward(*input_tensors)
..\..\bioimageio\core\prediction_pipeline\_model_adapters\_model_adapter.py:59: in forward
    return self._forward(*input_tensors)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <bioimageio.core.prediction_pipeline._model_adapters._torchscript_model_adapter.TorchscriptModelAdapter object at 0x00000251FDC143A0>
batch = (<xarray.DataArray (b: 1, c: 1, y: 256, x: 256)>
array([[[[-0.34453768, -0.5941573 , -0.85205287, ..., -0.5926006 ,
  ... , ..., -0.4026966 ,
[...]

PR #155, opened 2021-11-22
add unload for PredictionPipeline and ModelAdapters

FynnBe:
This PR
 - adds `load` and `unload` methods to `PredictionPipeline` and any `ModelAdapter` to occupy and free device (gpu) memory
 - enables the use of `PredictionPipeline` and any `ModelAdapter` as context
 - makes any `ModelAdapter` keep track if it's loaded and loads on the fly when calling forward and unloaded

With this PR the use of PredictionPipeline and ModelAdapter as context managers is encouraged, while the previous use is still fully functional.

Atm, only `pytorch_state_dict` and `pytorch_script` models actually unload, TF and keras model adapters give a warning and skip unloading.

### DIFF
--- a/bioimageio/core/prediction_pipeline/_prediction_pipeline.py
+++ b/bioimageio/core/prediction_pipeline/_prediction_pipeline.py
@@ -87,7 +87,6 @@ class _PredictionPipelineImpl(PredictionPipeline):
     def __init__(
         self, *, name: str, bioimageio_model: Model, processing: CombinedProcessing, model: ModelAdapter
     ) -> None:
-        super().__init__(bioimageio_model=bioimageio_model)
         if bioimageio_model.run_mode:
             raise NotImplementedError(f"Not yet implemented inference for run mode '{bioimageio_model.run_mode.name}'")
 
--- a/bioimageio/core/prediction_pipeline/_prediction_pipeline.py
+++ b/bioimageio/core/prediction_pipeline/_prediction_pipeline.py
@@ -23,7 +23,7 @@ class NamedImplicitOutputShape:
         return len(self.scale)
 
 
-class PredictionPipeline(ModelAdapter):
+class PredictionPipeline(abc.ABC):
     """
     Represents model computation including preprocessing and postprocessing
     """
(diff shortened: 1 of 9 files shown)


FynnBe:
note: tests need to be run (manually) with an available cuda device. torchscript tests fail for me, as on current main: #154
currently only single device management is tested

constantinpape:
We could also add a `close` method for cases where users can't use a context manager (e.g. prediction pipeline as member of a class)

bioimageio/core/prediction_pipeline/_prediction_pipeline.py
```diff
@@ -59,6 +68,20 @@ def output_specs(self) -> List[OutputTensor]:
...
+        """
+        ...
+
+    @abc.abstractmethod
+    def unload(self) -> None:
+        """
+        free any device memory in use
+        """
+        ...
+
```
FynnBe, line 84: there is basically a close() method here that I named unload() as load() makes more sense that open()...

bioimageio/core/prediction_pipeline/_prediction_pipeline.py
```diff
@@ -105,8 +139,11 @@ def postprocess(
         """Apply postprocessing."""
         return self._processing.apply_postprocessing(*input_tensors, input_sample_statistics=input_sample_statistics)
 
-    def __call__(self, *input_tensors: xr.DataArray) -> List[xr.DataArray]:
-        return self.forward(*input_tensors)
+    def load(self):
+        self._model.load()
+
+    def unload(self):
+        self._model.unload()
```
FynnBe, line 146: implementations of load() and unload() for PredictionPipeline

constantinpape:
> there is basically a close() method here that I named unload() as load() makes more sense that open()...

Ok, that also works :+1:

oeway:
I tried it a bit too with tensorflow, it was very tricky and I couldn't actually make it work. Have you fund some way to unload the model for tensorflow?

constantinpape:
> I tried it a bit too with tensorflow, it was very tricky and I couldn't actually make it work. Have you fund some way to unload the model for tensorflow?

We decided to focus on pytorch first and work on tf once we have a working version for pytorch.

PR #163, opened 2021-11-25
Fix shape val

constantinpape:
Fixes the input/output shape validation which did not work at all; we just did not notice because the tests for `test_model` where also broken...
In addition this also fixes the return codes from all CLI functions, which always returned 0 before.

### DIFF
--- a/bioimageio/core/__main__.py
+++ b/bioimageio/core/__main__.py
@@ -1,6 +1,7 @@
 import enum
 import json
 import os
+import sys
 from glob import glob
 
 from pathlib import Path
@@ -53,9 +54,10 @@ def package(
     # typer bug: typer returns empty tuple instead of None if weights_order_priority is not given
     weights_priority_order = weights_priority_order or None
 
-    return commands.package(
+    ret_code = commands.package(
         rdf_source=rdf_source, path=path, weights_priority_order=weights_priority_order, verbose=verbose
     )
+    sys.exit(ret_code)
 
 
 package.__doc__ = commands.package.__doc__
@@ -89,11 +91,12 @@ def test_model(
     )
     if summary["error"] is None:
         print(f"Model test for {model_rdf} has passed.")
-        return 0
+        ret_code = 0
     else:
         print(f"Model test for {model_rdf} has FAILED!")
         print(summary)
-        return 1
+        ret_code = 1
+    sys.exit(ret_code)
 
 
 test_model.__doc__ = resource_tests.test_model.__doc__
@@ -116,11 +119,12 @@ def test_resource(
     )
     if summary["error"] is None:
         print(f"Resource test for {rdf} has passed.")
-        return 0
+        ret_code = 0
     else:
         print(f"Resource test for {rdf} has FAILED!")
         print(summary)
-        return 1
+        ret_code = 1
+    sys.exit(ret_code)
 
 
 test_resource.__doc__ = resource_tests.test_resource.__doc__
@@ -159,7 +163,6 @@ def predict_image(
     prediction.predict_image(
         model_rdf, inputs, outputs, padding, tiling, None if weight_format is None else weight_format.value, devices
     )
-    return 0
 
 
 predict_image.__doc__ = prediction.predict_image.__doc__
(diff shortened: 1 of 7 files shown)


constantinpape:
Switching to the correct return codes has unearthed another bug: `test_cli_package[unet2d_nuclei_broad_model]` fails with

```
Traceback (most recent call last):                                                                                                                                                                                                                     
  File "/home/pape/Work/bioimageio/spec-bioimage-io/bioimageio/spec/shared/fields.py", line 425, in _deserialize                                                                                                                                       
    return raw_nodes.URI(uri_string=value)                                                                                                                                                                                                             
  File "<string>", line 9, in __init__                                                                                                                                                                                                                 
  File "/home/pape/Work/bioimageio/spec-bioimage-io/bioimageio/spec/shared/raw_nodes.py", line 109, in __post_init__                                                                                                                                   
    raise ValueError("Invalid URI or relative path. (use URI with scheme 'file' for absolute file paths)")                                                                                                                                             
ValueError: Invalid URI or relative path. (use URI with scheme 'file' for absolute file paths)                                                                                                                                                         
                                                                                                                                                                                                                                                       
The above exception was the direct cause of the following exception:                                                                                                                                                                                   
                                                                                                                                                                                                                                                       
Traceback (most recent call last):                                                                                                                                                                                                                     
  File "/home/pape/Work/bioimageio/python-bioimage-io/bioimageio/core/commands.py", line 36, in package                                                                                                                                                
    rdf_local_source = resolve_uri(rdf_source)                                                                                                                                                                                                         
  File "/home/pape/Work/software/conda/miniconda3/envs/bio-core-dev/lib/python3.7/functools.py", line 840, in wrapper                                                                                                                                  
    return dispatch(args[0].__class__)(*args, **kw)                                                                                                                                                                                                    
  File "/home/pape/Work/bioimageio/python-bioimage-io/bioimageio/core/resource_io/utils.py", line 161, in _resolve_uri_str                                                                                                                             
    return resolve_uri(fields.URI().deserialize(uri), root_path)                                                                                                                                                                                       
  File "/home/pape/Work/software/conda/miniconda3/envs/bio-core-dev/lib/python3.7/site-packages/marshmallow/fields.py", line 364, in deserialize                                                                                                       
    output = self._deserialize(value, attr, data, **kwargs)                                                                                                                                                                                            
  File "/home/pape/Work/bioimageio/spec-bioimage-io/bioimageio/spec/shared/fields.py", line 427, in _deserialize                                                                                                                                       
    raise ValidationError(value) from e                                                                                                                                                                                                                
marshmallow.exceptions.ValidationError: /tmp/bioimageio_cache/packages/UNet_2D_Nuclei_Broad_0_1_3p37.zip                                                                                                                                               
                                                                                                                                                                                                                                                       
During handling of the above exception, another exception occurred:          
```

This looks like an old problem that I hoped was fixed ....
@FynnBe I have no idea about this part, so I would suggest we merge this despite the test error (given that the CLI tests were just wrong beforehand) so that you can work on fixing this in a different PR (I am also not sure if this needs to be fixed here or in spec).

FynnBe:
This error stems from our mixed use of the URI classes (node and field) as paths and specifically that we allow relative paths, but no absolute ones. That we ended up with that has historic reasons only (e.g. we tried to avoid the `marshmallow_union` package)
To really clean this up we need to be strict about the use of URIs (a uri needs a scheme, thus any path is not a valid URI!) and use Union[URI, Path] where appropriate and enforce that paths are relative where they should be. (currently the URI just always raises on absolute paths which leads to these problems)

constantinpape:
Is this fixd by https://github.com/bioimage-io/spec-bioimage-io/pull/287 or do we need to change something else here?

constantinpape:
> Is this fixd by [bioimage-io/spec-bioimage-io#287](https://github.com/bioimage-io/spec-bioimage-io/pull/287) or do we need to change something else here?

No, looks like it still fails.

FynnBe:
> Is this fixd by [bioimage-io/spec-bioimage-io#287](https://github.com/bioimage-io/spec-bioimage-io/pull/287) or do we need to change something else here?

no, but hopefully by this WIP https://github.com/bioimage-io/spec-bioimage-io/pull/288 and its future equivalent here for bioimageio.core

PR #164, opened 2021-11-25
Uri vs path

FynnBe:
needs https://github.com/bioimage-io/spec-bioimage-io/pull/288

This PR updates the use of URI classes and pathlib.Path as relative paths are no longer considered valid URIs. Union[URI, Path] is used where appropriate.
Also improves some util functions.

### DIFF
--- a/bioimageio/core/build_spec/build_model.py
+++ b/bioimageio/core/build_spec/build_model.py
@@ -10,7 +10,7 @@ import numpy as np
 import bioimageio.spec as spec
 import bioimageio.spec.model as model_spec
 from bioimageio.core import export_resource_package, load_raw_resource_description
-from bioimageio.core.resource_io.utils import resolve_uri
+from bioimageio.core.resource_io.utils import resolve_source
 
 try:
     from typing import get_args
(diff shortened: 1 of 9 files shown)


constantinpape:
Similar to https://github.com/bioimage-io/spec-bioimage-io/pull/288 I can't really review this because the whole Path/URI management is a mess that I don't understand.
But looks like this fixes the tests now, so feel free to merge once you have made all the changes you want.

bioimageio/core/build_spec/build_model.py
```diff
@@ -1,7 +1,7 @@
 import datetime
 import hashlib
 import os
-from pathlib import Path
+from pathlib import Path, WindowsPath
```
constantinpape, line 4: WindowsPath is imported here but afaik not used in this file.

FynnBe: right, found a nice way (problem was the potential ":" in windows paths interfering with splitting dependency manager and file at ":"). cleaned it up

constantinpape:
Looks like zenodo is having timeout issues so the tests are failing ...

FynnBe:
> Looks like zenodo is having timeout issues so the tests are failing ...

I'd almost merge anyway given that 6b25a3b passes... let's give it another try first

constantinpape:
> I'd almost merge anyway given that [6b25a3b](https://github.com/bioimage-io/core-bioimage-io-python/commit/6b25a3b6c7f3367a73f5737709aff233b2f87e27) passes... let's give it another try first

Good call, now it passes :). I am gonna merge as soon as the conda build test is done.

PR #176, opened 2021-12-01
More tf features

### LINKED ISSUE #175: Better handling of tf version
We are not really making use of the tf version right now and should:
- raise a warning if the tf version is different than the one given by the model
- in `build_spec` require `tf_version` to be passed as argument instead of setting it to `1.15` as default value.
- do the same for opset_version for onnx

constantinpape:
Fixes #175

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -80,6 +80,7 @@ jobs:
       uses: conda-incubator/setup-miniconda@v2
       with:
         auto-update-conda: true
+        mamba-version: "*"
         channel-priority: strict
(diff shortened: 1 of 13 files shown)


constantinpape:
Also adds a weight converter for keras->tf weights.
Does not check the onnx version in prediction, I will create a separate issue on this.

constantinpape:
I had to change to mamba for the tf env, conda couldn't resolve it. I left it as conda for the other envs for now.

constantinpape:
all relevant tests pass, so I am gonna merge it.

PR #179, opened 2021-12-01
Add cover generation and default license to build_model, make input/o…

constantinpape:
…utput_axes mandatory


constantinpape:
Also moves `add_weigths` to its own file and adds tests for it.

PR #184, opened 2021-12-02
Check that samples are packaged in build_spec tests

constantinpape:
`sample_inputs` and `sample_outputs` are not automatically pacakged when included in the spec.
This should be fixed @FynnBe. (I added a test to `build_spec` this fails due to this issue; this causes the issue that Carlos just mentioned in the meeting.)


FynnBe:
needs https://github.com/bioimage-io/spec-bioimage-io/pull/319

constantinpape:
Tests on main all pass, so I am merging it.

PR #185, opened 2021-12-02
Test for issue with attachments

constantinpape:
@FynnBe there is still some issue with attachments: they are always empty in the serialized model and consequently the content of `attachments:files` is not copied. See also the comments in `build_model`.
This would be important to fix as the deepiamgej exporter relies on this to copy the pre/postprocessing macros!

(There is a test now that checks for this and currenttly fails.)


FynnBe:
main tests pass...

PR #193, opened 2021-12-13
Update prediction with tiling

constantinpape:
- [x] Add top-level imports for convenience
- [x] Add verbosity to prediction with tiling
- [ ] Enable prediction with tiling out of core


constantinpape:
> * Enable prediction with tiling out of core

This one is a bit more complex, and it's probably better to implement it in a separate function using dask. I will leave this for a later.

PR #207, opened 2022-01-22
Make sure unknown is not serialized in build_spec

constantinpape:


FynnBe:
the assert will always fail as the raw node always has the attribute "unknown", but unknown is no longer serialized to disk...
if you check the temporary yaml file:
```yaml
...
attachments:
  files: [zero_mean_unit_variance.ijm]
authors:
- {affiliation: EMBL Heidelberg, name: Constantin Pape;@bioimage-io}
...
```

PR #212, opened 2022-01-27
Change how the tiff sample data is written

constantinpape:


constantinpape:
Hey @esgomezm,
I have adapted your code from #212 here to write the sample tif data. I have exported a stardist model with the code here: https://oc.embl.de/index.php/s/EvU0nA04RT0GsHj. Could you please check if this can be correctly read?

If not we should discuss this quickly in person, that would be easier than going back and forth via mail or in the issue.

constantinpape:
Ok, this should work now.

PR #214, opened 2022-01-28
Build spec dep

constantinpape:


constantinpape:
Adds dependencies to all weight formats and also adds pytorch version. This PR also changes the signature of `build_model` a bit and removes the `**weight_kwargs` in favor of explicit arguments to avoid successfull function calls with unused keyword arguments.

This just touches `build_spec`, so I will merge as soon as all tests pass.

PR #224, opened 2022-02-17
Infer `.pt` weight file as `torchscript`

oeway:
`*.pt` file should be the extension for torchscript according to the documentation of pytorch: https://pytorch.org/tutorials/beginner/Intro_to_TorchScript_tutorial.html#saving-and-loading-models

### DIFF
--- a/bioimageio/core/build_spec/build_model.py
+++ b/bioimageio/core/build_spec/build_model.py
@@ -41,7 +41,9 @@ def _get_hash(path):
 
 def _infer_weight_type(path):
     ext = os.path.splitext(path)[-1]
-    if ext in (".pt", ".pth", ".torch"):
+    if ext == ".pt":
+        return "torchscript"
+    if ext in (".pth", ".torch"):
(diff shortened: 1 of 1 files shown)


constantinpape:
This makes sense, but breaks the test for our example models, which use `.pt` for pytorch state dict weights.
So we would need to either update these models, or update the respective tests here so that the weight type is not inferred from the extensions in those tests.

FynnBe:
This is generally problematic as `.pt` is also endorsed for pytorch model state dict: 

> A common PyTorch convention is to save models using either a .pt or .pth file extension.

https://pytorch.org/tutorials/beginner/saving_loading_models.html#:~:text=A%20common%20PyTorch%20convention%20is,that%20you%20must%20call%20model.

constantinpape:
Thanks for double checking @FynnBe. Then I would probably stick with the current solution where `weight_type=torchscript` has to be passed explicitly for torchscript weights.

PR #228, opened 2022-02-21
bump version

FynnBe:
and spec version


FynnBe:
... waiting for bioimageio.spec to be transfered from cf-staging to conda-forge channel

FynnBe:
@constantinpape should we maybe add the cf-staging channel in the CI to speed things up in these cases?

constantinpape:
> @constantinpape should we maybe add the cf-staging channel in the CI to speed things up in these cases?

I wouldn't do that, it may hide other issues. And it's already available according to https://anaconda.org/conda-forge/bioimageio.spec/files. This may be an issue with the CDN instead.

PR #229, opened 2022-02-24
Update build model interface

### LINKED ISSUE #211: keep interface of build_model in line with spec
probably a nitpick, but when building a model I ran into an unexpected type for the mandatory `cite` field/argument to build_model:

https://github.com/bioimage-io/core-bioimage-io-python/blob/902262bdce60a4c905f7b4a3b5646dc9ae68122f/bioimageio/core/build_spec/build_model.py#L528

[relevant code in spec](https://github.com/bioimage-io/spec-bioimage-io/blob/4c799d91293ec4785e7bb3e2e7a1313922e43921/bioimageio/spec/model/v0_4/schema.py#L282)

the spec states that this should be a list of dicts, but the build_model expects a single dict.

Otherwise, awesome how convenient building a model now is!

constantinpape:
Fixes #211

I think that this covers all the discrepancies between the build model interface and the spec, @oeway @k-dominik.
@oeway I also addressed your comment in https://github.com/bioimage-io/core-bioimage-io-python/issues/211#issuecomment-1042997222 a bit by changing the doc string to have "alt text" instead of saying "architecture".

I still need to update the examples in the notebook before merging this.


constantinpape:
Updating the notebook is pending fixing the example model, which needs https://github.com/bioimage-io/collection-bioimage-io/issues/294.

@k-dominik @oeway please check if the changes here are ok / if you think there are any other changes necessary to make `build_spec` fully reflect the spec.

example/bioimageio-core-usage.ipynb
```diff
@@ -60,41 +69,91 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 6,
```
k-dominik, line 72: I don't know if you want to keep outputs, but I'd at least run the notebook once from the beginning. Seems like cells executed 3-4 are lost somewhere, 5 seems to be out of order, this bears the danger that the output doesn't reflect linear flow through the notebook. Also there seem to be errors in the outputs.

constantinpape: I still need to fix the example model before the notebook can run again, but this is pending issues on the website... See also this comment: https://github.com/bioimage-io/core-bioimage-io-python/pull/229#issuecomment-1049908507.

k-dominik (commented):
awesome, will be even nicer to use now.

example/bioimageio-core-usage.ipynb
k-dominik: ah I see

oeway:
Hi, thanks for working on this!

I am thinking whether it would be better if we can run a quick static validation after calling this function (before creating the package), or have you already doing something like this?

constantinpape:
> I am thinking whether it would be better if we can run a quick static validation after calling this function (before creating the package), or have you already doing something like this?

This runs static validation by default.

oeway:
> > I am thinking whether it would be better if we can run a quick static validation after calling this function (before creating the package), or have you already doing something like this?
> 
> This runs static validation by default.

Awesome then! Looks good to me!

constantinpape:
Thanks for the feedback @oeway and @k-dominik. I will merge as soon as the issues with the example notebook are fixed.

bioimageio/core/build_spec/build_model.py
```diff
@@ -595,7 +602,7 @@ def build_model(
     authors: List[Dict[str, str]],
     tags: List[Union[str, Path]],
     documentation: Union[str, Path],
-    cite: Dict[str, str],
+    cite: List[Dict[str, str]],
     output_path: Union[str, Path],
     # model specific optional
     architecture: Optional[str] = None,
```
oeway, line 608: Would it make sense to deprecate the architecture and place it under a more generic weight kwargs field?

oeway:
Regarding the weights, is there a way to pass a static dict with several weights formats (instead of calling add_weights)?

constantinpape:
> Regarding the weights, is there a way to pass a static dict with several weights formats (instead of calling add_weights)?

Yes, we could change the weight format handling.
Can you make a concrete proposal for how the `build_model` interface should change?

constantinpape:
@oeway, after the feedback from @esgomezm in https://github.com/HenriquesLab/ZeroCostDL4Mic/pull/172 I would not introduce any more major changes to the function signature here.
Feel free to write up ideas about a better function signature in a separate issue, but we would only include this in a 0.6.0 release (next release will be 0.5.0, that will include the changes here and also add `training_data` to `build_model`).

constantinpape:
p.s. I noted that the way `parent` is passed is also not quite as in the spec.
However, the parent field definition seems quite outdated, see https://github.com/bioimage-io/spec-bioimage-io/issues/407.
I will bring this up in the meeting.

constantinpape:
I went ahead and also changed the parent signature.

PR #233, opened 2022-02-25
Add training data

constantinpape:
pending #229


constantinpape:
@FynnBe 
passing the training data as:
```
    training_data = { 
        "type": "dataset",
        "name": "nucleus-training-data",
        "description": "stardist nucleus training data",
        "source": "https://github.com/stardist/stardist/releases/download/0.1.0/dsb2018.zip",
    }   
```
fails with 
```
marshmallow.exceptions.ValidationError: {'training_data': ['Errors in all options for this field. Fix any of the following errors:', {'format_version': ['Missing data for required field.']}, {'type': ['Unknown field.'], 'source': ['Unknown fi
eld.'], 'description': ['Unknown field.'], 'name': ['Unknown field.']}]}
```
But I thought this should be supported. Could you clarify how to pass training data as a dataset rdf?

Using `training_data={"id": "..."}` works.

FynnBe:
let me try to dissect the error:
```
marshmallow.exceptions.ValidationError: {'training_data': 
['Errors in all options for this field. Fix any of the following errors:',   # this is from the Union; so there are multiple options, but all failed..
  {'format_version': ['Missing data for required field.']},   # option 1: missing "format_version" (to be a valid dataset RDF)
  {'type': ['Unknown field.'], 'source': ['Unknown field.'], 'description': ['Unknown field.'], 'name': ['Unknown field.']}   # option 2: anything that is not 'id' is unknown to the schema only expecting 'id'.
]}
``` 

the previously discussed `Union[str, Neted(DatasetRdf)]` would maybe give a somewhat clearer error...  ({...} not a valid string).

constantinpape:
Ah, I see; so it does need the format version.

constantinpape:
Is there a way how I can programatically get this? (I.e. the current format version for the dataset rdf? I don't want't to expose this since it's not something users should worry about.)

FynnBe:
> Is there a way how I can programatically get this? (I.e. the current format version for the dataset rdf? I don't want't to expose this since it's not something users should worry about.)

```
import bioimageio.spec

bioimageio.spec.dataset.format_version`
```
is probably what you are looking for

constantinpape:
Slightly unrelated: the issues with the marshmallow error messages is that it does not tell you which spec it is comparing to.
E.g. when you use `jsonschema` in python it actually prints the schema that is compared to, which makes the error longer, but also makes it possible to understand why it's not passing.

FynnBe:
> Slightly unrelated: one of the issues with the marshmallow error messages is that it does not tell you which spec it is comparing to. E.g. when you use `jsonschema` in python it actually prints the schema that is compared to, which makes the error longer, but also makes it possible to understand why it's not passing.

maybe (in general or just for a Union field) we can add a link to the documentation?

constantinpape:
> maybe (in general or just for a Union field) we can add a link to the documentation?
> e.g. https://github.com/bioimage-io/spec-bioimage-io/blob/gh-pages/model_spec_latest.md#format_version

Yeah, that would already be helpful.

PR #235, opened 2022-02-25
Bump minor version

constantinpape:
All the changes I would need are in now, anything else to add from your side, @FynnBe?


FynnBe:
there will always be another release for those...

PR #238, opened 2022-03-04
Add tests for build_model with parent field

constantinpape:
Indeed there are some spec changes necessary @FynnBe


FynnBe:
needs https://github.com/bioimage-io/spec-bioimage-io/pull/417

PR #240, opened 2022-03-07
add test_load_remote_rdf

FynnBe:
fixed by: https://github.com/bioimage-io/spec-bioimage-io/pull/419


constantinpape:
Looking good, the tests pass now on main. Let's wait till the spec release is on conda and then merge this once all tests pass.

constantinpape:
@FynnBe `test_load_remote_rdf` only works if pytorch is available. So we need to deactivate it if this is not the case.
I am not sure what the best solution with our current pytest config is, can you do it?

PR #241, opened 2022-03-10
Dataset statistics

FynnBe:
todo:
- [x] add tests
- [x] fix broken tests


bioimageio/core/measure_groups.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def reset(self):
+        self.n = 0
+        self.mean = None
+        self.m2 = None
+
+    def update_with_sample(self, sample: Dict[TensorName, xr.DataArray]):
+        tensor = sample[self.tensor_name]
+        mean_b = tensor.mean(dim=self.axes)
+        n_b = numpy.prod(tensor.shape) / numpy.prod(mean_b.shape)  # reduced voxel count
```
FynnBe, line 50: @constantinpape I'm not sure with the count/weight. Simply taking count 1 for 1 sample might neglect that samples may differ in the reduced dimensions. Thus here I take the voxel count as weight, but these may be big numbers, so I worry about numerical stability (see update formulas below)

bioimageio/core/measure_groups.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def reset(self):
+        self.count = 0
+        self.estimates = None
+
+    def update_with_sample(self, sample: Dict[TensorName, xr.DataArray]):
+        tensor = sample[self.tensor_name]
+        sample_estimates = tensor.quantile(self.qs, dim=self.axes)
+
+        n = numpy.prod(tensor.shape) / numpy.prod(sample_estimates.shape[1:])  # reduced voxel count
```
FynnBe, line 91: Analog voxel count as in MeanVarStd...

bioimageio/core/prediction_pipeline/_combined_processing.py
```diff
@@ -1,54 +1,30 @@
...
-    "scale_range": ScaleRange,
-    "sigmoid": Sigmoid,
-    "zero_mean_unit_variance": ZeroMeanUnitVariance,
-}
-
-
-Scope = Literal["sample", "dataset"]
-SAMPLE: Literal["sample"] = "sample"
-DATASET: Literal["dataset"] = "dataset"
-SCOPES: Set[Scope] = {SAMPLE, DATASET}
```
constantinpape, line 50: What happened to all the stuff that was removed here, was it moved somewhere else (if so I couldn't find that in the diffs), or is it irrelevant now.

constantinpape (commented):
The diff is too large for me to fully check now, I left some implementation comments though.
What's really missing from my side:
- tests
- usage examples

bioimageio/core/measure_groups.py
```diff
@@ -166,3 +169,29 @@ def finalize(self) -> Dict[Measure, MeasureValue]:
...
+                raise NotImplementedError(f"Computing datasets statistics for {m} not yet implemented")
+
+    measure_groups = []
+    for (tn, axes) in mean_var_std_groups:
+        measure_groups.append(MeanVarStd(tensor_name=tn, axes=axes))
+
+    for (tn, axes), ns in percentile_groups.items():
+        measure_groups.append(PercentileGroup(tensor_name=tn, axes=axes, ns=ns))
+
+    return measure_groups
```
FynnBe, line 197: a first version of how to map Measures to MeasureGroup

bioimageio/core/measure_groups.py
```diff
@@ -0,0 +1,168 @@
...
+
+try:
+    import crick
+except ImportError:
+    crick = None
+
+MeasureValue = xr.DataArray
+
+
+class MeasureGroup:
```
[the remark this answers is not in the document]
FynnBe: I added the docstring "group of measures for more efficient computation of multiple measures"
`MeanVarStd(MeasureGroup)` computes Mean, Var, and Std together, and the `...Percentiles(MeasureGroup)` computes multiple percentiles together.  Atm they can only be used for batched computation ('update_with_sample` + `finalize`). I am considering to add either separate groups for the sample case or using the same groups for simplicity (they should usually group just like the dataset case, although other algorithms might be used). 

The `Measure().compute` method could be used for reference (in the tests!) or (with a simple wrapper) as a trivial "group" (with a wrapper, because groups return Dict[Measure, MeasureValue], while `Measure` simply returns `MeasureValue`)

The mapping of a `Dict[TensorName, Set[Measure]]` to `List[MeasureGroup]` is done by `get_measure_groups`, see other comment.

bioimageio/core/measure_groups.py
```diff
@@ -0,0 +1,168 @@
...
+            self.estimates = (self.count * self.estimates + n * sample_estimates) / (self.count + n)
+
+        self.count += n
+
+    def finalize(self) -> Dict[Percentile, MeasureValue]:
+        assert self.estimates is not None
+        return {Percentile(n=n, axes=self.axes): e for n, e in zip(self.ns, self.estimates)}
+
+
+class TDigestPercentiles(MeasureGroup):
```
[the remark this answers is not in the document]
FynnBe: I think they have little overlap and it is cleaner to keep them separate. Maybe we add a third or throw out the crick one again... The separate setup should make testing easier as well, no?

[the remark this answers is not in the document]
FynnBe: I did rename it `CrickPercentiles` though 👍

bioimageio/core/measure_groups.py
FynnBe: I am now casting the input tensor to flaot64 and added a few asserts to ensure it stays there. For the percentiles I store the percentiles as float64 in `MeanPercentiles` although the error is probably big here anyway.... https://github.com/bioimage-io/core-bioimage-io-python/commit/a7c920669831190fde3c58b79746ccf3cd688f4f

bioimageio/core/measure_groups.py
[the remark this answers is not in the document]
constantinpape: > I think they have little overlap and it is cleaner to keep them separate. Maybe we add a third or throw out the crick one again... The separate setup should make testing easier as well, no?

It is an implementation detail users should not care about. But I will wait to see actual usage before I will comment more.

bioimageio/core/prediction_pipeline/_prediction_pipeline.py
```diff
@@ -147,20 +176,52 @@ def unload(self):
 
 
 def create_prediction_pipeline(
-    *, bioimageio_model: nodes.Model, devices: Optional[Sequence[str]] = None, weight_format: Optional[str] = None
+    *,
+    bioimageio_model: nodes.Model,
```
constantinpape, line 180: While we're working on the function signature here: can we allow `bioimageio_model` as positional argument? It is always annoying me that we have to write this full long kwarg and it's also not really in line with the rest of our functions.
(I am fine with not allowing positional for the rest as they are all optional)

constantinpape (commented):
These are a lot of changes and I don't really have the time to review this in depth now.
Overall I think that this looks ok, so we can go ahead with it. (Besides the small comment on `create_prediction_pipeline` function signature ;))
As some general comments (we don't need to address any of this in the current PR since it's already too large and for most of this using the current version first will also help):
- it would be very helpful to have some more concise usage examples, the notebook is pretty long and complex already.
- For the notebook some more structure and explanatory (markdown) text would help
- Maybe the usage could still be simplified, but I think we should go ahead and merge this now and test it first; some feedback from @k-dominik or others would also be helpful here.

PR #253, opened 2022-03-18
Fix torch to onnx conversion

constantinpape:
- remove torch version pinning to catch incompatibilities with newer versions
- Fix torch to onnx conversion, which failed due to https://github.com/pytorch/pytorch/pull/67809


FynnBe (approved):
LGTM

PR #256, opened 2022-04-01
Add functionality to overwrite dij config

constantinpape:


bioimageio/core/build_spec/util.py
```diff
@@ -0,0 +1,21 @@
...
+        dict_value_list = list()
+        for inner_dict in data_dict:
+            dict_value_list.append(replace_values(inner_dict, value_dict))
+        return dict_value_list
+    else:
+        for key in data_dict.keys():
+            value = data_dict[key]
+            new_value = value_dict.get(key, value)
+            if isinstance(value, dict) or isinstance(value, list):
+                new_dict[key] = replace_values(value, value_dict)
```
FynnBe, line 15: isn't it a bit weird that value_dict is not traversed into as well?

FynnBe: an example that confuses me: 
`orig_dict = {"a":  1, "b": {"a": 2}}`
`value_dict = {"a": 3}` vs `value_dict = {"b": {"a": 3}}`

constantinpape: I guess that would make it quite complicated. And it also would mean that you can't replace a simple value by a dict

FynnBe:
```suggestion
                new_dict[key] = replace_values(value, value_dict.get(key, {}))
```

bioimageio/core/build_spec/util.py
```diff
@@ -0,0 +1,21 @@
+
+
+# based on:
+# https://stackoverflow.com/questions/38491318/python-replace-keys-in-a-nested-dictionary
+def replace_values(data_dict, value_dict):
+    new_dict = {}
+    if isinstance(data_dict, list):
+        dict_value_list = list()
+        for inner_dict in data_dict:
+            dict_value_list.append(replace_values(inner_dict, value_dict))
```
FynnBe, line 8: in the list case the overwrite values should probably be a dict keyed by index... ?
```suggestion
        for i, inner_dict in enumerate(data_dict):
            dict_value_list.append(replace_values(inner_dict, value_dict.get(i, {})))
```

bioimageio/core/build_spec/util.py
FynnBe: > I guess that would make it quite complicated. And it also would mean that you can't replace a simple value by a dict

you can... your if else is on the data values... so this should work, no?
`orig_dict = {"a": 1}`
`value_dict = {"a": {"b": 2}}`

FynnBe: I suppose as it stands now you couldn't replace a dict with a simple value... so if you do want to go down this road we'd have to check if `value_dict` is actually a dict before traversing down...

FynnBe: the test case for that scenario would be
`orig_dict = {"a": {"b": 2}}`
`value_dict = {"a": 1}`

bioimageio/core/build_spec/util.py
constantinpape: This doesn't make sense. The conditional here should only depend on `data_dict` and not on `value_dict`.
`value_dict` should not be a list.

FynnBe: not a list, yes, but it could be a dict with index keys...
This is what I mean:
`orig_dict` = {"a": [0, 1, 2, 3]}`
`value_dict` = {"a": {3: 4}}`
`updated_dict` = {"a": [0, 1, 2, 4]}`

bioimageio/core/build_spec/util.py
constantinpape: Let's please not complicate this too much. Going into `value_dict` will introduce a lot of corner cases.
If you're not happy with the current solution I would drop the approach and instead pass a function to overwrite the deepij config. (This is a bit more complex for the user, but allows more flexibility.)

FynnBe: I think it would be helpful to have a generic rdf update function, which could also be used to update the rdfs with the content of the [resource.yaml](https://github.com/bioimage-io/collection-bioimage-io/blob/main/collection/10.5281/zenodo.5764892/resource.yaml). There, at the moment it is suboptimal to be able to only overwrite the root keys.
If we do add such a function it might be confusing if this part here behaves differently.
Either we 
1. never add such a general function (might still use such a function in the collection repo, but I think there is a more general use case for this)
2. maybe add a general update function at some point and to avoid confusing use a function as argument here
3. just add this general function now and use it here

I'm leaning towards option 3. as it seems to me the most user-friendly option, but option 2. would also be fine with me.

FynnBe: or option 4:
(general update function) using a flat update dict, but they keys are a key-path, e.g.
`orig_dict = {"a": 1, "b": {"a": 2}}`
`value_dict = {"b:a": 3}`

constantinpape: I think that if we want to implement something generic we should go for option 3 (and have this as a separate function).
Do you want to do this @FynnBe? We can then deprecate this PR.

FynnBe: ok, I'll add it to bioimageio.spec

PR #266, opened 2022-05-07
Refactor image functionality

constantinpape:
from prediction.py. In order to use this in dependencies, e.g. the BioEngine (though it will copy paste in  the beginning).

This is still lacking unittests for image_helper.py, will implement them later.


constantinpape:
Added some rudimentary tests now and gonna merge.

Issue #210, opened 2022-01-26
Save 3D, 2D+c or 2D+t test and sample images correctly


esgomezm:
Hi! 
I'm converting the ZCDL4M 3D U-Net notebook with the bioimageio.core code and I see that 3D images are not properly stored. This also happened to StarDist as far as I remember when the images were 2D + channels. Could it be possible to correct it?

Thank you!

constantinpape:
Hi @esgomezm,
I think that this is a discrepancy with how tifs with more than 2 dimensions are written here and read in deepImagej/Fiji.
And this is definitely something we want to fix. Currently, the code for writing the tif samples is [here](https://github.com/bioimage-io/core-bioimage-io-python/blob/main/bioimageio/core/build_spec/build_model.py#L357-L374).
So we would need to adapt this code so that it's compatible with deepIJ. But I don't really know how to test this easily, as I can read it back correctly in python, e.g. for the [stardist example model](https://oc.embl.de/index.php/s/k087NMVaH5PBNxU):
```python
import imageio
out = imageio.volread("sample_output_0.tif")
print(out.shape)
Out[3]: (512, 512, 33)
```

esgomezm:
Hi @constantinpape 
I answered you by email as well. But we can discuss it here. The code I normally use with tiff files is the following one:
```python
tifffile.imsave(file_name),
                    np.expand_dims(video, axis=[1, 2, -1]), # addapt the dimensions to tifffile ordering TZCYXS
                    resolution=(1 / pixel_size, 1 / pixel_size),
                    imagej=True,
                    metadata={'spacing': 1, 'unit': 'um'})
```
Storing the tiff like that shouldn't give you problems when recovering it from python. Could you test it?

Also, can use the very last version of the bioimageio.core in the notebook?

constantinpape:
Thanks for the code snippet @esgomezm. I am working on using it now, see #212.

> Also, can use the very last version of the bioimageio.core in the notebook?

Sorry, I don't quite understand what you mean, could you please clarify?

constantinpape:
@esgomezm I think I have fixed the issues now and I have made a new release already that includes the changes: v0.4.10. If you switch to it in the zero cost notebooks the problems should be fixed.

esgomezm:
Hi!

[...]

PR #275, opened 2022-06-01
Correct the axes that create sample tiffs

carlosuc3m:
Correct errors converting npy test images into images in tiff format, https://github.com/bioimage-io/core-bioimage-io-python/issues/210

### DIFF
--- a/bioimageio/core/build_spec/build_model.py
+++ b/bioimageio/core/build_spec/build_model.py
@@ -420,10 +420,10 @@ def _write_sample_data(input_paths, output_paths, input_axes, output_axes, pixel
         # convert the image to expects (Z)CYX axis order
         if im.ndim == 4:
             assert set(axes) == {"b", "x", "y", "c"}, f"{axes}"
-            axes_ij = "cyxb"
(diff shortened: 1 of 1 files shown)


constantinpape:
Hi @carlosuc3m,
thanks for working on this. But your changes break the tests:
```
        if rgb is None:
            rgb = shape[-1] in (3, 4) and ndim > 2
        if rgb and shape[-1] not in (3, 4):
            raise ValueError('ImageJ hyperstack is not a RGB image')
        if not rgb and ndim == 6 and shape[-1] != 1:
>           raise ValueError('ImageJ hyperstack is not a grayscale image')
E           ValueError: ImageJ hyperstack is not a grayscale image
```
So it looks like your new solution is also not correct yet.

carlosuc3m:
Hello again @constantinpape all test seems to pass now except the one for tf-legacy which seems to fail while building the dependencies. Do you think this is related to the code added?
Regards,
Carlos

constantinpape (approved):
Thanks @carlosuc3m, looks good now!

Issue #278, opened 2022-06-12
create_prediction_pipeline from RawResourceDescription


oeway:
Hi, with @k-dominik we are working on creating a triton model adapter which implements its own model execution logic. For that we would like to bypass the model weights downloading and prefer to use `load_raw_resource_description` to provide a  `RawResourceDescription` to the `create_prediction_pipeline`. @FynnBe or @constantinpape could you please provide some hint on how to do this, currently it fails.

```
 File "/models/bioengine-model-runner/1/model.py", line 147, in execute
E             pred_pipeline = create_prediction_pipeline(
E           File "/tmp/python_env_g4jJUf/0/lib/python3.8/site-packages/bioimageio/core/prediction_pipeline/_prediction_pipeline.py", line 200, in create_prediction_pipeline
E             preprocessing = CombinedProcessing(bioimageio_model.inputs)
E           File "/tmp/python_env_g4jJUf/0/lib/python3.8/site-packages/bioimageio/core/prediction_pipeline/_combined_processing.py", line 33, in __init__
E             raise NotImplementedError(t)
E         NotImplementedError: InputTensor(name='input0', data_type='uint8', axes='bcyx', shape=ParametrizedInputShape(min=[1, 1, 64, 64], step=[0, 0, 16, 16]), preprocessing=[Preprocessing(name='zero_mean_unit_variance', kwargs={'axes': 'cyx', 'mode': 'per_sample'})], description=<marshmallow.missing>, data_range=(0.0, 255.0))
```

constantinpape:
Sounds good. I don't have time to look into this now, can you take a look @FynnBe?

FynnBe:
I will look into this. How urgent is it?

oeway:
> I will look into this. How urgent is it?

Would be great if we have it soon, I made some attempt but not working. Basically I need it for a much better implementation of the bioengine model runner. In this new implementation, we let triton to handle all the models, and the bioimageio.core only do pre/postprocessing, that can greatly improve the throughput of the server.

oeway:
@FynnBe Any updates on this? We have some stability issue with the bioengine which I hope we can fix with this new implementation.

FynnBe:
> @FynnBe Any updates on this? We have some stability issue with the bioengine which I hope we can fix with this new implementation.

Hi @oeway yes, I finally got around to it! 
[...]

PR #284, opened 2022-07-18
allow to create prediction pipeline from raw_nodes.Model

FynnBe:
implements https://github.com/bioimage-io/core-bioimage-io-python/issues/278


oeway:
Great! Thanks!

PR #299, opened 2022-10-24
basic support for scaling during tiled prediction

pattonw:
I've added a super simple model for testing. Tiled prediction fails for all of the existing models (some assumptions broken like too many inputs/outputs, some predictions not quite exactly equal to result so maybe the halo's aren't quite right)
Not entirely sure how to add a model to the `raw.githubusercontent.com/bioimage-io/spec-bioimage-io/main/example_specs/models` folder, or what would be appropriate to add, for now all the files I used are included here.


bioimageio/core/prediction.py
```diff
@@ -342,7 +362,8 @@ def predict_with_tiling(
             if any(sc != 1 for ax, sc in scale.items() if ax in "xyz") or any(
                 off != 0 for ax, off in offset.items() if ax in "xyz"
             ):
-                raise NotImplementedError("Tiling with a different output shape is not yet supported")
+                pass
```
constantinpape, line 365: I think we can just remove this whole check, including the `if`.

bioimageio/core/prediction.py
```diff
@@ -270,8 +279,10 @@ def check_tiling(tiling):
         spatial_axes = [ax for ax in axes if ax in "xyz"]
         halo = tiling["halo"]
         tile = tiling["tile"]
+        scale = tiling.get("scale", dict())
         assert all(halo.get(ax, 0) >= 0 for ax in spatial_axes)
         assert all(tile.get(ax, 0) > 0 for ax in spatial_axes)
+        assert all(scale.get(ax, 1.0) > 0 for ax in spatial_axes)
```
constantinpape, line 285: Can you get `output.scale` here as default instead of `1.0`? Then this would also yield correct results for cases where the scale changes and where users specify custom parameters for halo and tile (but not for scale).

bioimageio/core/prediction.py
```diff
@@ -288,6 +299,14 @@ def check_tiling(tiling):
             assert isinstance(shape, list)
             assert len(shape) == len(axes)
 
+            scale = None
+            output_shape = output_spec.shape
+            if not isinstance(output_shape, list):
+                scale = output_shape.scale
+            else:
+                scale = [1.0] * len(axes)
```
constantinpape, line 307: In this case we have a hard-coded output shape. I think it's not clear that this would correspond to having `scale=1.0`.
I would rather throw a NotImplementedError here.

tests/conftest.py
```diff
@@ -63,6 +75,9 @@
         "https://raw.githubusercontent.com/bioimage-io/spec-bioimage-io/main/example_specs/models/"
         "stardist_example_model/rdf_wrong_shape2.yaml"
     ),
+    "shape_change": (
+        "/home/skar/Work/Code/Packages/core-bioimage-io-python/shape_change_model.yaml"
```
constantinpape, line 79: Can you move the model to `example_specs/models` and replace the local filepath with the url?

constantinpape (changes requested):
Hi @pattonw,
thanks for working on this! I left a few comments on the code changes, but this part looks good overall.
 I would suggest to split up this PR into two parts though: i) make a new PR that just adds your sample PR to `models/example_specs` (without code changes), once it's merged you can then directly refer to the correct url in `conftest.py` and the tests can run correctly.

bioimageio/core/prediction.py
pattonw: Done!

bioimageio/core/prediction.py
pattonw: hmm, I changed this a bit more. It seemed awkward to have `scale.get(ax, output_spec.shape.scale.get(ax, 1)) > 0`. Definitely doable if that's the right approach. Instead I ensured that any metadata provided in output_spec is read into the tiling dict first, and then overwritten with whatever the users provides, which should achieve the same thing. Let me know if you think that more consequences than just fixing this issue as now a user could provide a custom "halo", "tile", or "scale" term and in all cases the anything not provided will fall back to the defaults that would have been generated on a `tiling=True`

bioimageio/core/prediction.py
pattonw: Fixed. I put an assert at the top so that tiling will always fail if the output_shape is provided as a simple list rather than an implicit shape. This is a breaking change in the cases where scale was not provided, but input_shape and output_shape were, with output_shape being a simple list.
It makes sense to me that tiling should not work if the output shape is simply provided as a list.

tests/conftest.py
pattonw: I don't see an `example_specs/models` directory, is that in another repo?

FynnBe: the examples are in the spec folder where you'd have to make a separate PR..
https://github.com/bioimage-io/spec-bioimage-io/tree/main/example_specs/models

constantinpape: Thanks for clarifying @FynnBe. Indeed, the examples would need to go into the other repository.

tests/conftest.py
```diff
@@ -23,6 +24,17 @@
 keras_models = ["unet2d_keras"]
 tensorflow_js_models = []
 
+
```
constantinpape, line 27: This looks like a left-over from debugging?

tests/conftest.py
pattonw: :thumbsup: Thanks!

bioimageio/core/prediction.py
```diff
@@ -263,43 +278,59 @@ def _parse_tiling(tiling, input_specs, output_specs):
...
-
-            tiling = {
-                "halo": {ax: ha for ax, ha in zip(axes, halo) if ax in "xyz"},
-                "tile": {ax: sh for ax, sh in zip(axes, shape) if ax in "xyz"},
-            }
-            check_tiling(tiling)
-        else:
-            tiling = None
+    elif isinstance(tiling, bool) and not tiling:
+        tiling = None
```
constantinpape, line 333: What does actually happen in this case? The whole function interface is currently a bit awkward, because calling the function with `tiling=False` doesn't make sense. This is due to the implementation of some higher level functions that make use of this function and others, so I don't want to change it right now. But we may want to raise a ValueError here instead of setting tiling to None.

bioimageio/core/prediction.py
constantinpape: Yes, I think that's much better.

constantinpape (commented):
@pattonw the changes look good! If you change the example data as discussed this is good to go I think (though I will do one more round of reading the changes more carfeully).

tests/prediction_pipeline/test_prediction_pipeline_tiled.py
```diff
@@ -0,0 +1,51 @@
+import numpy as np
```
constantinpape, line 1: We actually have tests for the tiled prediction already: https://github.com/bioimage-io/core-bioimage-io-python/blob/main/tests/test_prediction.py#L138
I don't mind refactoring these tests to the file here.

tests/prediction_pipeline/test_prediction_pipeline_tiled.py
```diff
@@ -0,0 +1,51 @@
...
+        xr.DataArray(np.load(str(test_tensor)), dims=tuple(spec.axes))
+        for test_tensor, spec in zip(bio_model.test_outputs, bio_model.outputs)
+    ]
+    assert len(outputs) == len(expected_outputs)
+
+    for out, exp in zip(outputs, expected_outputs):
+        assert_array_almost_equal(out, exp, decimal=4)
+
+
+def test_prediction_pipeline_torch(any_torch_model):
```
constantinpape, line 34: I think we can't run these tests with `any...` model yet, because we have some models with multiple outputs etc. that are not supported by the tiled prediction yet.
Best to refactor the tests linked to above to here and extend them by the shape change model.

bioimageio/core/prediction.py
pattonw: Seems like it will just throw an indexing error later when you try to access tiling["tile"]. I'll change it to throw a value error with a more helpful error message than just "None type does not support indexing"

pattonw: Nevermind, there's already a ValueError if calling `predict_with_tiling` and `tiling=False`

pattonw: replaced with an error just to make it obvious to future devs

tests/prediction_pipeline/test_prediction_pipeline_tiled.py
pattonw: Refactored my tests into `test_prediction.py` and made the tests specific with respect to the model. I added 2 models to the tiling tests that both pass, including my recently added simple test model.

On that note, this pull request breaks the existing `test_predict_image_with_tiling_fixed_output_shape` test since we now are more strict about throwing an error when trying to tile with fixed output shape.

constantinpape: Thanks for the changes.

> On that note, this pull request breaks the existing `test_predict_image_with_tiling_fixed_output_shape` test since we now are more strict about throwing an error when trying to tile with fixed output shape.

Ok, I will think about that case tomorrow (a bit late here for this right now ;)).

constantinpape: I think we need to enable the case of a explicit output shape if the input shape is also explicit and if they are identical. I made a small PR with these changes, see below.

constantinpape:
@pattonw I made a small PR onto this one that fixes a typo in the new test url and that adds a special case for explcit input and output shapes that are identical (this is potentially a very common case that we should cover): https://github.com/pattonw/core-bioimage-io-python/pull/1.

pattonw:
seems like a fairly safe special case to allow

constantinpape:
I think all tests should pass now and this is be good to go from my side. Anything else you want to add @pattonw or should I merge?

pattonw:
No, I think its all good to go

constantinpape (approved):
Awesome, thanks for the contribution :).

PR #304, opened 2022-10-30
Add support for tensorflow 2

### LINKED ISSUE #302: Weight conversion for tensorflow 2 is not yet implemented.
I'm trying to  export a TensorFlow 2 model into deepImageJ format but it gives errors as the function seems to not be implemented for it:

```python
from bioimageio.core.weight_converter.keras import convert_weights_to_tensorflow_saved_model_bundle
```
```python
WARNING:tensorflow:Compiled the loaded model, but the compiled metrics have yet to be built. `model.compile_metrics` will be empty until you train or evaluate the model.
1/1 [==============================] - 0s 153ms/step
---------------------------------------------------------------------------
NotImplementedError                       Traceback (most recent call last)
[<ipython-input-22-95f300b64ebb>](https://localhost:8080/#) in <module>
    197 if os.path.exists(tf_weight_path):
    198   rmtree(tf_weight_path)
--> 199 convert_weights_to_tensorflow_saved_model_bundle(output_path, tf_weight_path + ".zip")
    200 add_weights(output_path, tf_weight_path + ".zip", output_path, tensorflow_version=tf.__version__)
    201 

[/usr/local/lib/python3.7/dist-packages/bioimageio/core/weight_converter/keras/tensorflow.py](https://localhost:8080/#) in convert_weights_to_tensorflow_saved_model_bundle(model_spec, output_path)
    105         return _convert_tf1(weight_path, str(path_), model.inputs[0].name, model.outputs[0].name, zip_weights)
    106     else:
--> 107         raise NotImplementedError("Weight conversion for tensorflow 2 is not yet implemented.")

NotImplementedError: Weight conversion for tensorflow 2 is not yet implemented.
```
An example of the notebook to reproduce this error can be checked here: https://github.com/HenriquesLab/ZeroCostDL4Mic/pull/219

The notebook can also be used with binary images but just in case, data from [here](https://zenodo.org/record/5639253#.Y1vuMuzP06E) can also be used.
@constantinpape

constantinpape:
Fixes #302

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -69,6 +69,33 @@ jobs:
       shell: bash -l {0}
       run: pytest --disable-pytest-warnings
   
+  test-base-bioimage-spec-tf:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.7, 3.8, 3.9]
+    steps:
+    - uses: actions/checkout@v2
+    - name: install dependencies
+      uses: conda-incubator/setup-miniconda@v2
+      with:
+        auto-update-conda: true
+        # we need mamba to resolve environment-tf
+        mamba-version: "*"
+        channel-priority: flexible
+        activate-environment: bio-core-tf
+        environment-file: dev/environment-tf.yaml
+        python-version: ${{ matrix.python-version }}
+    - name: additional setup
+      shell: bash -l {0}
+      run: |
+        conda remove --force bioimageio.spec
+        pip install --no-deps git+https://github.com/bioimage-io/spec-bioimage-io
+        pip install --no-deps -e .
+    - name: pytest-base-bioimage-spec-tf
+      shell: bash -l {0}
+      run: pytest --disable-pytest-warnings
+
   test-base-bioimage-spec-tf-legacy:
     runs-on: ubuntu-latest
     strategy:
@@ -83,8 +110,8 @@ jobs:
         # we need mamba to resolve environment-tf
         mamba-version: "*"
         channel-priority: flexible
-        activate-environment: bio-core-tf
-        environment-file: dev/environment-tf.yaml
+        activate-environment: bio-core-tf-legacy
+        environment-file: dev/environment-tf-legacy.yaml
         python-version: ${{ matrix.python-version }}
     - name: additional setup
       shell: bash -l {0}
--- a/bioimageio/core/prediction_pipeline/_model_adapters/_tensorflow_model_adapter.py
+++ b/bioimageio/core/prediction_pipeline/_model_adapters/_tensorflow_model_adapter.py
@@ -97,9 +97,14 @@ class TensorflowModelAdapterBase(ModelAdapter):
 
         return res
 
-    def _forward_keras(self, input_tensors):
+    def _forward_keras(self, *input_tensors):
         tf_tensor = [tf.convert_to_tensor(ipt) for ipt in input_tensors]
-        result = self._model.forward(*tf_tensor)
+
+        try:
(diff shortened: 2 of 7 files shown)


constantinpape:
@FynnBe could you have a look at the failing tests?
I think the problem is that we also need to distinguish between keras models that are compatible with tf1 / tf2. But I don't quite understand the logic in conftest for doing this.

esgomezm:
> @FynnBe could you have a look at the failing tests? I think the problem is that we also need to distinguish between keras models that are compatible with tf1 / tf2. But I don't quite understand the logic in conftest for doing this.

Example of a keras model here: https://github.com/bioimage-io/core-bioimage-io-python/issues/302#issuecomment-1296977109

FynnBe:
errors in [31ff72f](https://github.com/bioimage-io/core-bioimage-io-python/pull/304/commits/31ff72f34fca0c48ac6b71fda95dfed420a16a39) might have been caused by some http error, leading to an emtpy response, throwing a json error... could not reproduce this locally...

constantinpape:
> Example of a keras model here: [#302 (comment)](https://github.com/bioimage-io/core-bioimage-io-python/issues/302#issuecomment-1296977109)

I have added an example model based on https://bioimage.io/#/?tags=bacillus-subtilis&id=10.5281%2Fzenodo.7261974 already @esgomezm.

> errors in [31ff72f](https://github.com/bioimage-io/core-bioimage-io-python/pull/304/commits/31ff72f34fca0c48ac6b71fda95dfed420a16a39) might have been caused by some http error, leading to an emtpy response, throwing a json error... could not reproduce this locally...

Thanks for taking care of the tests @FynnBe.
I can't reproduce the test failures locally either, even after clearing the cache. Not sure why this is going wrong consistently in the tests here. I will run again in a few hours to see if this is temporary.

constantinpape:
I get this test failure locally though:
```
__________________________________________________________________________________ test_assert_dtype __________________________________________________________________________________
                                                                                                                                                                                       
    def test_assert_dtype():                                                                                                                                                           
        from bioimageio.core.prediction_pipeline._processing import AssertDtype                                                                                                        
                                                                                                                                                                                       
        proc = AssertDtype("test_tensor", dtype="uint8")                                                                                                                               
        tensor = xr.DataArray(np.zeros((1,), dtype="uint8"), dims=tuple("c"))                                                                                                          
        out = proc(tensor)                                                                                                                                                             
        assert out is tensor                                                                                                                                                           
                                                                                                                                                                                       
        tensor = tensor.astype("uint16")                                                                                                                                               
        with pytest.raises(AssertionError):                                                                                                                                            
            out = proc(tensor)                                                                                                                                                         
>           assert out is tensor                                                                                                                                                       
E           Failed: DID NOT RAISE <class 'AssertionError'>                                                                                                                             
                                                                                                                                                                                       
prediction_pipeline/test_processing.py:27: Failed  
```

But let's first see if we get the other tests to run.

FynnBe:
> I get this test failure locally though:

this is really confusing. I just double checked this one locally for me and it passes... looking at it I have no idea how that one could fail... ?!?
hmm... maybe if you somehow have some optimization flags active that skip the assert statement in `AssertDtype`?

FynnBe:
interestingly the error for py3.7 is: `AttributeError: 'str' object has no attribute 'decode'`

constantinpape:
These test errors are very weird... Let's see if they go away tomorrow...

FynnBe:
the CI installed tensorflow 2.4.1 which is older than what was used to create the test model.. (2.9)...

constantinpape:
Awesome, thanks for figuring it out @FynnBe.

PR #306, opened 2022-11-03
Update setup.py

FynnBe:
we ought to pin bioimageio.spec more closely due to lack of forward compatibility. 
i.e. installing an older version of core will install the newest spec version.

This will add to the list of checkboxes for a new core release, but should make the released versions more reliable.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ setup(
     ],
     packages=find_namespace_packages(exclude=["tests"]),  # Required
     install_requires=[
-        "bioimageio.spec>=0.4.5",
+        "bioimageio.spec==0.4.8",
         "imageio>=2.5",
         "numpy",
         "ruamel.yaml",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -26,7 +26,7 @@
     ],
     packages=find_namespace_packages(exclude=["tests"]),  # Required
     install_requires=[
-        "bioimageio.spec>=0.4.5",
+        "bioimageio.spec==0.4.8",
```
constantinpape, line 29: I think that pin is too hard since it will only allow to install exact matches to `0.4.8`, but not `0.4.8post1` etc.
See https://www.darius.page/pipdev/ for checking pip dependencies.
I think `==0.4.8.*` should be fine.

FynnBe:
```suggestion
        "bioimageio.spec==0.4.8.*",
```

constantinpape (approved):
LGTM

PR #310, opened 2022-11-04
Workflow RDF

FynnBe:
needs https://github.com/bioimage-io/spec-bioimage-io/pull/478

Implements resolved nodes for the workflow RDF.


oeway (commented):
Looks good! One suggestion, for running workflow, would be nice to support running it step by step via an python iterator.

For example, for debugging a workflow, the user of the core library can do something like this:


for step in workflow:
    print(workflow.context) # print the current available context, e.g. previous outputs, workflow RDF source
    print(step.inputs, step.options)
    step.execute() # run the step, can also use try except to capture error
    print(step.outputs)

What do you think?

FynnBe:
> What do you think?

not sure if this really provides value. For debugging one can simply add a breakpoint in the for loop of the run_workflow op... Not sure it's a good idea to basically recommend debugging with print...
If we want this we can of course add this, but I wouldn't conside thist urgent, as we are not sure how to proceed from this draft yet.

oeway:
Well, think about a user use our package without access to the source code (which is the case for running in triton server). 

This is not about debugging with print, you can use pdb the same way. Also, keep in mind that the run_workflow is inside our package. Vscode for example cannot add breakpoint for pip packages, but it can add breakpoint to user's own code. Even if we can add breakpoints for packages, not everyone uses an IDE. We don't want to ask everyone to clone our repo and tap into our source code, just for debugging a workflow.

For triton inference server, we pack the core into a conda package, and it runs inside a docker container, no pdb, no ide, print is the only way to debug. I would say this would be an important feature practically. Especially in the beginning, we want to know what happens when running the workflow.

FynnBe:
You raise some good points I didn't think of yesterday. I'll look into implementing the workflow steps with `yield`... 👍

bioimageio/core/workflow/operators/_run.py
```diff
@@ -0,0 +1,202 @@
...
+    named_outputs: Dict[str, Any]
+
+
+def iterate_workflow_steps(
+    rdf_source: Union[dict, PathLike, IO, str, bytes, raw_nodes.URI, RawResourceDescription],
+    *,
+    inputs: Sequence = tuple(),
+    options: Dict[str, Any] = None,
+) -> Generator[WorkflowState]:
+    yield from _iterate_workflow_steps_impl(rdf_source, inputs=inputs, options=options, test_steps=False)
```
FynnBe, line 82: @oeway would this separate op suffice?
If the behavior only depends on a flag the return type of `run_workflow` changes somewhat drastically from an output tuple to a generator (of intermediate output tuples or--like I implemented now--a state object).

FynnBe:
most content---workflow implementations and workflow exectuion logic--was moved to https://github.com/bioimage-io/workflows-bioimage-io-python
the resolved nodes remain.

FynnBe:
unfortunately I mixed in the renaming of `ImportedSource` to `ImportedCallable` into this PR and it's counterpart in spec: https://github.com/bioimage-io/spec-bioimage-io/pull/478

It's just some renaming to avoid confusion about source as in source data and source as in a python callable defined in a module or provided source code.

FynnBe:
together with new repo https://github.com/bioimage-io/workflows-bioimage-io-python this updated PR supersedes #312 and #316

PR #318, opened 2022-11-24
MyPy check

FynnBe:
While updating type annotations for bioimageio.spec in https://github.com/bioimage-io/spec-bioimage-io/pull/489 I started working on type checking core as well...


FynnBe:
we are using pyright these days

Issue #311, opened 2022-11-09
Support adding multiple weights using build_model


oeway:
Discussion from the BMZ meeting, proposed by @ivan-ea 

Would be nice to support adding multiple weights in the build_model function.

cc @constantinpape @FynnBe

ivan-ea:
For context:
I was not sure if it was possible to generate models with multiple weight files with the build_model function.
Also I wondered if I missed something, since the website has the option to choose between several weight formats to download.

Some models have 2 weights in the rdf (for example [impartial-shrimp](https://bioimage.io/#/?id=10.5281%2Fzenodo.5874741)), so I was curious how they might have built them.

ivan-ea:
Thanks @constantinpape  the add_weights function seems exactly what I was missing.
I was following the model creation notebook, since I need to get the
deepimagej_config too. Maybe I could add the add_weights to the example notebook
once I figure out how to use it.

constantinpape:
> Maybe I could add the add_weights to the example notebook
> once I figure out how to use it.

That would be great!

PR #321, opened 2022-12-05
add_weights example in model_creation notebook

ivan-ea:
Adress issue #311

Example of use of the `add_weights` function in the model_creation notebook. I modified the 3rd example.

It takes an original model with `pytorch_state_dict weights`, builds the rdf.yaml with the deepimagej configuration, 
and then adds the information of the `torchscript` weights. 
Finally, it tests to see that both weight types produce the same results.


constantinpape (approved):
Thanks @ivan-ea!

Issue #319, opened 2022-11-30
Improve TF Model Adapter


FynnBe:
We should improve the TF Model Adapter to be more robust.

I can't improve this right now, but here are some notes from looking into it briefly, if someone else wants to tackle it already:

We should improve the TF Moder adapter here: https://github.com/bioimage-io/core-bioimage-io-python/blob/bc25746cad127e6a54d0885c2212aa8f50a049f8/bioimageio/core/prediction_pipeline/_model_adapters/_tensorflow_model_adapter.py#L65-L66

here `weight_file` is a folder in which we should look for the expected saved_model.pb file and possibly add the nested folder to its path.

We potentially unzip nested zip files (calling self.require_unzipped twice, which we probably shouldn't)

We also implicitly assume the use of tensorflow 1: https://github.com/bioimage-io/core-bioimage-io-python/blob/bc25746cad127e6a54d0885c2212aa8f50a049f8/bioimageio/core/prediction_pipeline/_model_adapters/_tensorflow_model_adapter.py#L34-L35

Maybe https://www.tensorflow.org/api_docs/python/tf/saved_model/save is a relevant reference for updating this part.

_Originally posted by @FynnBe in https://github.com/bioimage-io/core-bioimage-io-python/issues/317#issuecomment-1331904114_

PR #322, opened 2023-01-20
TF: prevent model compilation

jdeschamps:
Hi!

I ran into an issue with N2V: since it uses a custom loss, loading the model using `tf.keras.models.load_model` throws an error when the model has been compiled and saved as a `SavedModel`. The same method is used when validating a `SaveModel` in `bioimageio.core`.

One way around this is to pass `compile=False` ([see that line](https://github.com/jdeschamps/core-bioimage-io-python/blob/d2505c47a826ec68a1576eab56172efd9f947459/bioimageio/core/prediction_pipeline/_model_adapters/_tensorflow_model_adapter.py#L32)), avoiding the compilation of the model after loading. If I am not mistaken, the compilation is only required for training (compiling loss, metrics, optimizer etc.), it is not needed for the purpose of the bioimage.io model validation (inference). 

The model can still be loaded and compiled by users provided they have access to the custom methods. I tested that with N2V and it solves the error.

Happy to know more if there are things that I overlooked!

### DIFF
--- a/bioimageio/core/prediction_pipeline/_model_adapters/_tensorflow_model_adapter.py
+++ b/bioimageio/core/prediction_pipeline/_model_adapters/_tensorflow_model_adapter.py
@@ -29,7 +29,7 @@ class TensorflowModelAdapterBase(ModelAdapter):
     def _load_model(self, weight_file):
         weight_file = self.require_unzipped(weight_file)
         if self.use_keras_api:
-            return tf.keras.models.load_model(weight_file)
+            return tf.keras.models.load_model(weight_file, compile=False)
         else:
             # NOTE in tf1 the model needs to be loaded inside of the session, so we cannot preload the model
             return str(weight_file)


esgomezm:
Hey! It's the same with some of the ZeroCostDL4Mic notebooks. So yes, you would need to export it without compiling it, and then, when loading it into the N2V framework, you could also load the weights only into a compiled architecture.

jdeschamps:
> Hey! It's the same with some of the ZeroCostDL4Mic notebooks. So yes, you would need to export it without compiling it, and then, when loading it into the N2V framework, you could also load the weights only into a compiled architecture.

Yes, within the n2v that's also not really an issue because can we can load it with `custom_objects`.

But since there is no need to compile the model to run the inference for the model validation in bioimage.io (as far as I can tell), I thought it would make more sense for the model validation to avoid this error, hence the PR. Although this is an edge case, that lifts slightly the constraints on the uploaders' side.

Now, maybe that's not a good choice because it would then force users/consumers to also load with the `compile=False` parameter. In which case, maybe there should also be a TF example for saving models (e.g. [here](https://github.com/bioimage-io/core-bioimage-io-python/tree/main/example)) to help people avoid the issue?

Finally, the CI seems to fail on my PR during dependency installation for TF and TF-legacy. Any clue why? Out of curiosity. :)

Feel free to close the PR :)

constantinpape:
I am not familiar enough with keras to see the implications of this right away. It would be good to have the CI going over this, since this could potentially be an impactful change. Unfortunately our legacy tf 1 environment is not working again...
Maybe it's time to drop that one in favor of a maintainable tf 2 one?? cc @FynnBe

FynnBe:
While not too familiar with keras, I see no issue with adding the `compile=False` flag on loading 👍 
CI should pass though, which looks unrealdated (somewhow in the tf 3.7 env py 3.10 is present...).

FynnBe:
relates to #319

FynnBe (approved):
LGTM
let's wait for https://github.com/bioimage-io/core-bioimage-io-python/pull/323 and have CI pass here...

PR #336, opened 2023-05-14
Add multichan output model to test issues in build-spec

constantinpape:

PR #337, opened 2023-05-16
Fix tests

FynnBe:


constantinpape:
Just fyi: you can avoid the onnx related bugs with pytorch 2.0 by adding `onnx` to the env, see my changes in https://github.com/bioimage-io/core-bioimage-io-python/pull/336

constantinpape:
And I am not sure if keeping the `test-spec-rf-legacy` env around is sustainable; if you can't fix it I suggest we get rid of it.

FynnBe:
> Just fyi: you can avoid the onnx related bugs with pytorch 2.0 by adding `onnx` to the env, see my changes in #336

ah, thanks!

FynnBe:
> And I am not sure if keeping the `test-spec-rf-legacy` env around is sustainable; if you can't fix it I suggest we get rid of it.

I hope https://github.com/bioimage-io/core-bioimage-io-python/pull/337/commits/29f4b77511224b039e6007c37bb779cb8e2e3835 has fixed the legacy specific issue...

FynnBe:
hmm.. so, the main issue was the caching. I'll try to get it work properly to speed up testing.

constantinpape:
> hmm.. so, the main issue was the caching. I'll try to get it work properly to speed up testing.

Ok, can we do this in a separate PR though? I think we should release the fix for the grid image as soon as possible, it's currently blocking some other releases for me + it means that `build_model` is not working for users who have a newer PIL / imageio version.

FynnBe:
yes, of course! I'll merge once the tests ran through to make sure the tests really pass now and do the caching separately! 👍

PR #340, opened 2023-05-16
Check file integrity with SHA256 as part of any resource test

FynnBe:
needs https://github.com/bioimage-io/spec-bioimage-io/pull/508

### DIFF
--- a/bioimageio/core/resource_io/utils.py
+++ b/bioimageio/core/resource_io/utils.py
@@ -7,6 +7,8 @@ import sys
 import typing
 from types import ModuleType
 
+from marshmallow import missing
+
 from bioimageio.spec.shared import get_resolved_source_path, raw_nodes, resolve_source, source_available
 from bioimageio.spec.shared.node_transformer import (
     GenericRawNode,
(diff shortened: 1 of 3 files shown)


bioimageio/core/resource_io/utils.py
```diff
@@ -54,6 +57,68 @@ def generic_visit(self, node):
...
+                else:
+                    continue
+
+                if not hasattr(node, source_name):
+                    raise ValueError(
+                        f"Node {node} expected to have '{source_name}' field associated with '{expected_sha256}'"
+                    )
+
+                source_node = getattr(node, source_name)
+                source = get_resolved_source_path(source_node, root_path=self.root_path)
```
k-dominik, line 104: not a super strong opinion at all, but I think readability here could be improved. `field`, `expected_sha256` are more like key, value, and only at after all the checking we're sure what we have. It would be great to have "correct" meaning from the start- so I'd probably factor out a generator that only returns fields that are `sha256`-like fields.

k-dominik (approved):
lgtm

PR #359, opened 2023-11-14
Fix critical bug in predict with tiling

constantinpape:
This fixes a critical bug where the values passed for tiling are not taken into account and the code always uses the default tiling. This is very nasty, because the default tiling is usually quite small and leads to blocking artifacts.


FynnBe:
can't detect any obvious systematic differences:
exp: exp
out: out

Going to allow higher deviation to pass the currently failing test `test_predict_image_with_tiling_2`

PR #378, opened 2024-04-24
Fix pytorch_state_dict import from library

jdeschamps:
## Issue

`PyTorchModelAdaptor` passes a `sha256` keyword to the callable responsible for instantiating the model:
https://github.com/bioimage-io/core-bioimage-io-python/blob/ff8af21f51907d27a4cc2f6103249d22d40947b7/bioimageio/core/model_adapters/_pytorch_model_adapter.py#L107

If the callable is an `ArchitectureFromLibraryDescr`, then an error will be raised since there is no `kwargs` in the function signature:
https://github.com/bioimage-io/core-bioimage-io-python/blob/ff8af21f51907d27a4cc2f6103249d22d40947b7/bioimageio/core/digest_spec.py#L64

## Fix

Add `kwargs` to the function signature (and that of other callables).
https://github.com/bioimage-io/core-bioimage-io-python/blob/7681ab7a6c10774e200a8d4b6ebf75dbc3a36453/bioimageio/core/digest_spec.py#L65

## SIde note

The pre-commit hooks forced me to fix an import, is it due to me not having the right version of the specs?
https://github.com/bioimage-io/core-bioimage-io-python/blob/7681ab7a6c10774e200a8d4b6ebf75dbc3a36453/bioimageio/core/digest_spec.py#L23

### DIFF
--- a/bioimageio/core/digest_spec.py
+++ b/bioimageio/core/digest_spec.py
@@ -20,7 +20,7 @@ from typing import (
 from numpy.typing import NDArray
 from typing_extensions import Unpack, assert_never
 
-from bioimageio.spec._internal.io_utils import HashKwargs, download
+from bioimageio.spec._internal.io import HashKwargs, download
 from bioimageio.spec.common import FileSource
 from bioimageio.spec.model import AnyModelDescr, v0_4, v0_5
 from bioimageio.spec.model.v0_4 import CallableFromDepencency, CallableFromFile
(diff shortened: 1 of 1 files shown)


FynnBe:
> The pre-commit hooks forced me to fix an import, is it due to me not having the right version of the specs?

no, not sure why this wasn't picked up before.. I suppose it was a ruff check? the fix is good anyway.

FynnBe:
nice fix!
sad that `singledispatch` does play nice with static type checking... we should probably remove it in favor of some if else switch to have this part statically checked.

PR #395, opened 2024-06-13
adding example bioimageio.core_usage notebook and requirements txt file

mariana-gferreira:


FynnBe:
Hi @mariana-gferreira 
as your notebook is an updated version of https://github.com/bioimage-io/core-bioimage-io-python/blob/main/example/model_usage.ipynb
maybe we shouldn't add it as a modified copy, but rather keep the name.. 
that would make reviewing this PR also easier (as then I'd see the diff right away).

mariana-gferreira:
@FynnBe Done! 😄

example/model_usage.ipynb
```diff
@@ -2,132 +2,333 @@
...
+    "for out in model.outputs:\n",
+    "    print(f\"\\noutput '{out.id}' with axes:\")\n",
+    "    pprint(out.axes)\n",
+    "    print(f\"Data description: {out.data}\")\n",
+    "    print(f\"Test tensor available at:  {out.test_tensor.source.absolute()}\")\n",
+    "    if len(out.postprocessing) > 1:\n",
+    "        print(\"This output is postprocessed with: \")\n",
+    "        for p in out.postprocessing:\n",
+    "            print(p)"
+   ]
```
FynnBe, line 295: These parts are copied from https://github.com/bioimage-io/spec-bioimage-io/blob/main/example/load_model_and_create_your_own.ipynb
we should avoid this. If it is unclear that the model description can be inspected like that we can reference the spec example notebook here again

example/model_usage_requirements.txt
FynnBe, line 1: could be replaced as mentioned above..

FynnBe (commented):
I know this PR has been merged already, but I only had the time now to take a closer look and noticed some minor things that maybe could be improved. What do you think @esgomezm @mariana-gferreira ?

example/model_usage_requirements.txt
thodkatz: Regarding `model_requirements.txt` inspired by this [scanpy tutorial](https://github.com/scverse/scanpy-tutorials/blob/02c4946e0be47e033355ef84b2a4909b302d2513/scanpy_workshop/day1_01_GColab.ipynb), we can include the packages listed in the `requirements.txt`, within a cell of the notebook. This should be convenient for anyone that just opens only the notebook in the gogle collab or even locally, without cloning the repo, and creating dev environments.

e.g. in the first cell of the notebook
```python
%pip install bioimageio.core==0.6.7 torch==2.3.1 onnxruntime==1.18.0 matplotlib == 3.9.0
```

, and we can remove the check `if os.getenv("COLAB_RELEASE_TAG")`, since even if it is local or google colab, we need the same dependencies. From a dev perspective, we can update the conda's environment file and `setup.py`.

Also, the python version listed in the current `model_requirements.txt` can't be installed with pip e.g. `pip install -r requirements.txt` will fail. Conda can install python interpreter though. But in this use case, we can assume that we have python installed, and we want to install our dependencies, using only the notebook.

example/model_usage.ipynb
thodkatz: The inspection of the model is indeed referenced to the spec tutorial. Maybe both `1.3 Inpsect model data`, and `1.4 Get expected model inputs and outputs` could be delegated to spec. For this tutorial, we shouldn't focus on the inspection, but we could have a quick summary of the model such as with `pprint(model)`? Would be nice maybe for spec to use `tabulate` package to have a nice table overview (e.g. authors, description, git repo, ...)

example/model_usage.ipynb
```diff
@@ -2,132 +2,333 @@
...
+    "    print(f\"\\nThe model '{model.name}' with URL '{BMZ_MODEL_URL}' has been correctly loaded.\")\n",
+    "else:\n",
+    "    print('\\nPlease specify a model ID, DOI or URL')\n",
+    "\n",
+    "if \"draft\" in BMZ_MODEL_ID or \"draft\" in BMZ_MODEL_DOI or \"draft\" in BMZ_MODEL_URL:\n",
+    "    print(f\"\\nThis is the DRAFT version of '{model.name}'. \\nDraft versions have not been reviewed by the Bioimage Model Zoo Team and may contain harmful code. Run with caution.\")\n",
+    "\n",
+    "# To be added later:\n",
+    "# elif model.version != model.lastest_version:\n",
+    "#   print('\\nThe loaded version of the model is: ' + model.version, 'but the latest version of the model is: ' + model.lastest_version)\n",
```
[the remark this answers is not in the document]
thodkatz: Do you recommnend to just delete the commented code, or implement the second option?

example/model_usage.ipynb
```diff
@@ -2,132 +2,333 @@
...
+    "# Load general dependencies\n",
+    "from imageio.v2 import imread\n",
+    "from bioimageio.spec.utils import download\n",
+    "from pprint import pprint\n",
     "import matplotlib.pyplot as plt\n",
-    "from numpy.typing import NDArray\n",
+    "import numpy as np\n",
     "\n",
-    "try:\n",
-    "    import napari\n",
```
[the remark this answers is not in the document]
thodkatz: For napari, since in our example, we have 2D images and matplotlib seems to do the job, we could just keep it simple?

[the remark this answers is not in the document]
FynnBe: sure, we can keep it simple 👍

example/model_usage.ipynb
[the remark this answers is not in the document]
FynnBe: we should remove this whole section.
Loading from different sources is covered in https://github.com/bioimage-io/spec-bioimage-io/blob/a078cd44a48c7e9bf2cbb7c7b58bf903b5b834c6/example/load_model_and_create_your_own.ipynb#L107-L145

example/model_usage.ipynb
FynnBe: yes, we should refer to the spec notebook instead of repeating so much.

An overview could be nice, for this purpose I also considered https://python-devtools.helpmanual.io/usage/#prettier-print , 
but I don't think we need to add that dependency to spec. 
We could add it [here](https://github.com/bioimage-io/core-bioimage-io-python/blob/3179f83cda85351f1ff4c00639580ffe9d0e32b6/setup.py#L51), but it does make this notebook a bit more convoluted. 

Maybe for a canonical rendering of a description we could simply render the YAML file, or use https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html#IPython.display.JSON to let jupyter display the metadata.
Simply printing `f"loaded {model.name}"` would do for now I suppose.

example/model_usage_requirements.txt
FynnBe: yes, but I'm not a fan of just having pip install commands in the notebook when running it locally.. 
it may be executed unintentionally and mess with the local env.
maybe we do
```python
try:
    import bioimage.core
except ImportError:
    %pip install ...
    import bioimageio.core
    
```
?
so one can still use "exec all cells (above)" without up-/downgrading the local env.

example/model_usage.ipynb
thodkatz: yep actually the pprint, seems an easy solution :) Something like this should be fine
image

example/model_usage_requirements.txt
thodkatz: That's a nice idea as well! So that way we can still handle the use case of having everything in one for other users that just want the notebook, but also for devs that have already an environment installed.

On the other hand in theory, the dependencies declared in the notebook should be compatible with the dev environment. A problematic use case would be to have a dev environment different than the one defined in the notebook. So I guess that a new environment only for the notebook is needed eventually to be sure, and for the developer to test it that works for users.

But anyway, I think we are overthinking it :) I think your proposed change makes sense.

example/model_usage.ipynb
FynnBe: not sure what you mean...
- the pprint from devtools would handle pydantic models
- to avoid more dependencies we should probably just have a quick look at https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html#IPython.display.JSON , might be particularly suited for this situation (in a notebook)

we should not get hung up on this though. Especially anything not using another lib is totally fine without further consideration

thodkatz: The above `pprint` is native python implementation, so no external dependencies. I just wanted to show, that it nicely renders the model's info. Not as pretty as it would be with pprint from devtools :)

For IPython.display.JSON couldn't make it work.

Issue #398, opened 2024-07-08
Further improve example notebook


FynnBe:
A starting point to improve the example notebook(s) further is to digest the post-merge review of #395

cc @esgomezm

PR #401, opened 2024-07-11
Fix prediction with blocking

thodkatz:
## Issue

Using the jupyter notebook [model_usage.ipynb](https://github.com/bioimage-io/core-bioimage-io-python/blob/main/example/model_usage.ipynb), if we attempt to test the functionality of `predict_sample_with_blocking` instead of the `predict_sample_without_blocking` in the corresponding cell of the 3.2 section, there seems to exist some tiling artifacts:

```python
prediction: Sample = prediction_pipeline.predict_sample_with_blocking(sample)

# show the prediction result
show_images(sample, prediction)
```

Output:
output

It should be noted that the `predict_sample_without_blocking` it has expected behavior.

## Solution

The problem lies in the definition of the `local_slice` in the `bioimageio/core/block_meta.py`.
```python
    def local_slice(self) -> PerAxis[SliceInfo]:
        """inner slice wrt the block, **not** the sample"""
        return Frozen(
            {
                a: SliceInfo(
                    self.padding[a].left,
                    self.padding[a].left + self.inner_shape[a],
                )
                for a in self.inner_slice
            }
        )
```
Instead of using the padding to adjust the `SliceInfo`, the halo needs to be used. Padding is applied only when the expansion needs to include pixels outside of the actual image. So the fix was very simple:
```python
    def local_slice(self) -> PerAxis[SliceInfo]:
        """inner slice wrt the block, **not** the sample"""
        return Frozen(
            {
                a: SliceInfo(
                    self.halo[a].left,
                    self.halo[a].left + self.inner_shape[a],
                )
                for a in self.inner_slice
            }
        )
```

The interesting thing though about this issue, isn't the solution, but the thought process of converging to the problem.

Tiling artifacts as it has been discussed is a quite hot topic, and as it seems there are many factors involved, when we see something that goes off in the segmentation when we do blocking. The problems could have to do with:

- Normalization of the blocks. Having a block that displays only the background, and with a normalization that is local instead of having the global view of the image, can result for the model to pick up the noise of the block, and can create some artifacts
- The blocking process: a) halo, b) padding, c) cropping, d) merging

To help the process of trying to understand what could be the issue, I tried to expose from the `prediction_with_blocking` the input blocks, and the predicted ones adding a `verbose` option. To do that I have introduced two classes the `PredictionSampleMetadata`, and the `PredictionTensorMetadata`, that expose the input and predicted blocks.

Then from the jupyter notebook, we can visualize them, and we can actually have the following:

```python
prediction: PredictionSampleMetadata = prediction_pipeline.predict_sample_with_blocking(sample, verbose=True)
show_block_prediction(member_idx=0)
```
image

This visualization shows what the model took as input and how it behaves. That was very convenient when I was tying to debug the issue, because, you can interpret for example if the inputs are correct, and if the model behaves accordingly. So in our case the input blocks and the predicted ones were fine, but the final merged predicted image was off. So we can navigate that it was a merging issue.

I could also improve the visualization to display some lines to indicate where we do the padding.

## ... But

As I was experimenting the parameterized size, we have the following outputs:

```python
prediction: PredictionSampleMetadata = prediction_pipeline.predict_sample_with_blocking(sample, ns=3, verbose=True)
show_block_prediction(member_idx=0)
```
image

Seems okey, but when we have this

```python
prediction: PredictionSampleMetadata = prediction_pipeline.predict_sample_with_blocking(sample, ns=1, verbose=True)
show_block_prediction(member_idx=0)
```

image

We can see that the input blocks fed to the model seems very noisy. So the problem starts from the input itself. I would like to inverstigate this more on why this is happening.


thodkatz:
That could be also connected with ilastik issue regarding maybe normalization https://github.com/ilastik/ilastik/issues/2701

bioimageio/core/_prediction_pipeline.py
```diff
@@ -40,6 +43,52 @@
...
+        assert len(self.input_blocks) == len(self.prediction_blocks)
+        assert is_perfect_square(self.total_blocks)
+
+    @property
+    def total_blocks(self):
+        return len(self.input_blocks)
+
+
+@dataclass(frozen=True)
+class PredictionSampleMetadata:
```
FynnBe, line 63: not entirely sure about the purpose of this class.. I'll leave a series of comments and maybe then we can have a chat about it later.

1. I see how you want a structure to capture the relation of an input to an output. However,
    -  having both `input: Sample` and `input_blocks: List[SampleBlock]` seems redundant. (fine for debugging though)
    - how does `SampleBlockMeta` fit into this?
2. `Meta` in the name suggests to me that there is no actual data (but there is, this object has `input: Sample`)

bioimageio/core/_prediction_pipeline.py
```diff
@@ -40,6 +43,52 @@
...
+    def __post_init__(self):
+        assert len(self.input.members) == len(self.prediction.members)
+        assert len(self.input_blocks) == len(self.prediction_blocks)
+        assert is_perfect_square(self.total_blocks)
+
+    @property
+    def total_blocks(self) -> int:
+        return len(self.input_blocks)
+
+    def get_prediction_tensor_metadata(self, idx: int) -> PredictionTensorMetadata:
```
FynnBe, line 78: The rest of the code references sample members by their `MemberId`. Would be great to keep that consistent

bioimageio/core/block_meta.py
```diff
@@ -161,8 +161,8 @@ def local_slice(self) -> PerAxis[SliceInfo]:
         return Frozen(
             {
                 a: SliceInfo(
-                    self.padding[a].left,
-                    self.padding[a].left + self.inner_shape[a],
+                    self.halo[a].left,
+                    self.halo[a].left + self.inner_shape[a],
                 )
```
FynnBe, line 166: this is the actual fix!! I'm glad you found! Maybe we make a sparate PR for it if we need more time incorporating your new verbose features?

bioimageio/core/common.py
```diff
@@ -125,3 +126,7 @@ class SliceInfo(NamedTuple):
 
 #     def __repr__(self) -> str:
 #         return f"{type(self).__name__}({self.mapping!r})"
+
+def is_perfect_square(num: int) -> bool:
+    square = np.sqrt(num)
+    return np.ceil(square) == np.floor(square)
```
FynnBe, line 132:
```suggestion
    return np.ceil(square) == np.floor(square)

```

example/model_usage.ipynb
```diff
@@ -503,7 +503,7 @@
    "source": [
     "Use the new prediction pipeline to run a prediction for the previously loaded test image.\n",
     "\n",
-    "The prediction pipeline returns a `Sample` object, which will be displayed via napari."
+    "The prediction pipeline returns a `Sample` object, which will be displayed."
```
FynnBe, line 506: This relates to #398
Maybe we move changes to the model_usage notebook to yet another PR to adress some of the outstanding issues with it.

tests/test_bioimageio_spec_version.py
```diff
@@ -29,8 +29,8 @@ def test_bioimageio_spec_version(mamba_cmd: Optional[str]):
     # get currently pinned bioimageio.spec version
     meta = metadata("bioimageio.core")
     req = meta["Requires-Dist"]
-    assert req.startswith("bioimageio.spec ==")
-    spec_ver = req[len("bioimageio.spec ==") :]
+    assert req.startswith("bioimageio.spec==")
```
FynnBe, line 32: why does this need to change?
there is a space in https://github.com/bioimage-io/core-bioimage-io-python/blob/66665c573b06a43ae5a0e17d81961619235a66b2/setup.py#L32

FynnBe (changes requested):
Great writeup and awesome that you found the `halo` vs `padding` bug!
Let's maybe have a quick chat on how to proceed...

tests/test_bioimageio_spec_version.py
thodkatz: When I cloned the repo and run the tests, it was the only test that was failing due to this space. Maybe I am missing something. Also I am not sure why the CI don't trigger the tests for this PR.

FynnBe: interesting. I enable CI now for this PR

bioimageio/core/_prediction_pipeline.py
thodkatz: Yep, actually for the classes, I am not sure if they should be there. I just used them for  debugging, but maybe they are not a fit for the package. I just wanted a placeholder to check input, outputs, and their respective blocks. They could be easily deleted and just keep the bug fix.

The Metadata I think of it as a way of going beyond just the prediction output, and check the whole process, of what was the input, and the blocks. Maybe not the prefect name. Could be `PredictionSampleContainer`?

thodkatz:
so, as we have discussed with @FynnBe, first of all this PR, is a a bit mixed up, because it has a small bug fix, but at the same time introduces this logic of the `verbose` option, that needs more discussion about the visualization. So for this PR, we could keep only the bugfix. I will push a new commit only with the bugfix.

FynnBe (approved):

PR #406, opened 2024-07-23
add predict command

FynnBe:
This PR
- updates the CLI (rewrite that now include a predict command)
- includes some other changes to allow loading statistical measures from json with pydantic validation


thodkatz:
This is very cool :) I have already tested with affable-shark and the test input data provided by the model.

I have one general question, maybe I am missing something. Since we can have multiple input tensors and output tensors, that means we can have multiple `.npy` files. Why we don't store the tensors in one `.npy` file as an object, or maybe use the `.npz`, so we can can make our lifes easier dealing with one input and one output file?

FynnBe:
I used `.npy` as a file format default here, but in general we support many file formats. Users might want use `.h5`, `.tif` `.png`... which the imageio library takes care of for us.

As there is no `.npz` equivalent across file formats I would like to keep this simple approach. 
We might want to use `.h5` as a default and we should add `.zarr` and possibly make that the default instead.
Using numpy is the most hacker friendly, but not the most user friendly.

thodkatz:
Oh I see, maybe we could add this to the help section, so users can know what are the possible format right? For example, I didn't know that also from the format of the output file, the format of the file is inferred as well e.g. if --outputs=*.png, a png image will be created.

Althoug file extensions and file formats can be out of sync, but yeah let's not be too pedantic :)

bioimageio/core/cli.py
```diff
@@ -0,0 +1,425 @@
...
+    stat = stats_calc.finalize()
+    _ = stats_path.write_bytes(stat_adapter.dump_json(stat))
+
+    return stat
+
+
+class PredictCmd(CmdBase, WithSource):
+    """bioimageio-predict - Run inference on your data with a bioimage.io model."""
+
+    inputs: Union[str, Sequence[str]] = "model_inputs/*/{tensor_id}.*"
```
thodkatz, line 187: How do you pass a list of inputs?

I have attempted to use
`bioimageio predict affable-shark --inputs input1.npy input2.npy`, but the `input2.npy` isn't recognized. We need argparse for that no?

bioimageio/core/cli.py
```diff
@@ -0,0 +1,425 @@
...
+        n_glob_matches: Dict[int, List[str]] = {}
+
+        if isinstance(self.inputs, str):
+            if len(input_ids) > 1 and "{tensor_id}" not in self.inputs:
+                raise ValueError(
+                    f"{self.descr_id} needs inputs {input_ids}. Include '{{tensor_id}}' in `inputs` or provide multiple input paths/glob patterns."
+                )
+
+            inputs = [self.inputs.replace("{tensor_id}", t) for t in input_ids]
+        else:
```
thodkatz, line 238: Not very important, but usually with if else statements, I try to make it more assertive

```python
if isinstance(self.inputs, str):
...
elif isinstance(self.inputs, list):
...
else:
    raise ValueError(f"Not expected type {self.inputs}")
```

README.md
```diff
@@ -9,6 +9,113 @@
 
 Python specific core utilities for bioimage.io resources (in particular models).
 
+## Get started
+
+To get started we recommend installing bioimageio.core with conda together with a deep
+learning framework, e.g. pytorch, and run a few `bioimageio` commands to see what
+bioimage.core offers.
+
+1. install with conda (for more details on conda environments, [checkout the ])
```
thodkatz, line 18: just a minor heads up to fill this link :)

bioimageio/core/cli.py
FynnBe: good question, we should probably document that...
https://docs.pydantic.dev/latest/concepts/pydantic_settings/#lists

- `--inputs='[input1.npy,input2.npy']`
- `--inputs input1.npy --inputs input2.npy`
- `--inputs input1.npy,input2.npy`

PR #408, opened 2024-07-31
Update _torchscript_model_adapter.py

esgomezm:
converting the input to the model to float to avoid numpy array errors from torch

@oeway


FynnBe (changes requested):
should not be hardcoded IMO

oeway:
I agree it should not be directly converted to float32, however, it should not leave it as is either since pytorch does not support all numpy types, it seems all the unsigned types cause error, so we can either do float32 right now, or explicitly write a mapping for unsupported types.

FynnBe:
But `data.type` is a required field. So there is always a known, fixed data type at this point in the code.
(field `data_type` for v0_4 models).
Could you please point me to the example where you encountered the issue?

PR #418, opened 2024-08-29
Use relative and absolute tolerance for validating model outputs

### LINKED ISSUE #416: Validate model output with precision relative to the values of the data
### Problem

Some datasets have large values, so when validating that an output matches the given expected output, it doesn't always make sense to use decimal precision as a metric. Say for example a dataset has values on the order `1e5` then the output could match the expected output to 5 significant figures but still fail the validation test, even if the decimals are checked only up to the first position past the decimal point.

### Solution

It would be desirable to be able to specify something like the number of significant figures or some kind of relative tolerance to validate the outputs with.

melisande-c:
Changed model output validation to use relative and absolute tolerance instead of decimal precision. Using relative tolerance makes more sense because it is a more consistent measure of closeness — the arguments don't have to be changed if the data values are large or small. This was implemented by replacing `numpy.testing.assert_almost_equal` with `numpy.testing.assert_allclose`. Legacy support has been included for the `decimal` parameter, if it is used it will cause the validation to revert to the previous behaviour.

Resolves #416

### DIFF
--- a/bioimageio/core/_resource_tests.py
+++ b/bioimageio/core/_resource_tests.py
@@ -1,8 +1,9 @@
 import traceback
 import warnings
 from itertools import product
-from typing import Dict, Hashable, List, Literal, Optional, Set, Tuple, Union
+from typing import Dict, Hashable, List, Literal, Optional, Set, Tuple, Union, TypedDict
 
+from attr import dataclass
 import numpy as np
 from loguru import logger
 
@@ -35,14 +36,21 @@ def test_model(
     source: Union[v0_5.ModelDescr, PermissiveFileSource],
     weight_format: Optional[WeightsFormat] = None,
     devices: Optional[List[str]] = None,
-    decimal: int = 4,
+    absolute_tolerance: float = 1e-4,
+    relative_tolerance: float = 1e-1,
+    decimal: Optional[int] = None,
 ) -> ValidationSummary:
     """Test model inference"""
+    precision_args = _transform_precision_args(
+        absolute_tolerance=absolute_tolerance,
+        relative_tolerance=relative_tolerance,
+        decimal=decimal
+    )
     return test_description(
         source,
         weight_format=weight_format,
         devices=devices,
-        decimal=decimal,
+        **precision_args,
         expected_type="model",
     )
 
@@ -53,16 +61,23 @@ def test_description(
     format_version: Union[Literal["discover", "latest"], str] = "discover",
     weight_format: Optional[WeightsFormat] = None,
     devices: Optional[List[str]] = None,
-    decimal: int = 4,
+    absolute_tolerance: float = 1e-4,
+    relative_tolerance: float = 1e-1,
+    decimal: Optional[int] = None,
     expected_type: Optional[str] = None,
 ) -> ValidationSummary:
     """Test a bioimage.io resource dynamically, e.g. prediction of test tensors for models"""
+    precision_args = _transform_precision_args(
+        absolute_tolerance=absolute_tolerance,
+        relative_tolerance=relative_tolerance,
+        decimal=decimal
+    )
     rd = load_description_and_test(
         source,
         format_version=format_version,
         weight_format=weight_format,
         devices=devices,
-        decimal=decimal,
+        **precision_args,
         expected_type=expected_type,
     )
     return rd.validation_summary
@@ -74,10 +89,18 @@ def load_description_and_test(
     format_version: Union[Literal["discover", "latest"], str] = "discover",
     weight_format: Optional[WeightsFormat] = None,
     devices: Optional[List[str]] = None,
-    decimal: int = 4,
+    absolute_tolerance: float = 1e-4,
+    relative_tolerance: float = 1e-1,
+    decimal: Optional[int] = None,
     expected_type: Optional[str] = None,
 ) -> Union[ResourceDescr, InvalidDescr]:
     """Test RDF dynamically, e.g. model inference of test inputs"""
+    precision_args = _transform_precision_args(
+        absolute_tolerance=absolute_tolerance,
+        relative_tolerance=relative_tolerance,
+        decimal=decimal
+    )
+
     if (
         isinstance(source, ResourceDescrBase)
         and format_version != "discover"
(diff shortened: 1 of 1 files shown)


bioimageio/core/_resource_tests.py
```diff
@@ -1,8 +1,9 @@
 import traceback
 import warnings
 from itertools import product
-from typing import Dict, Hashable, List, Literal, Optional, Set, Tuple, Union
+from typing import Dict, Hashable, List, Literal, Optional, Set, Tuple, Union, TypedDict
 
+from attr import dataclass
```
FynnBe, line 6: seems to be unused?

bioimageio/core/_resource_tests.py
```diff
@@ -361,6 +397,50 @@ def _test_expected_resource_type(
...
+        + "`relative_tolerance` and `absolute_tolerance`, with different validation "
+        + "logic, using `numpy.testing.assert_allclose, see "
+        + "'https://numpy.org/doc/stable/reference/generated/"
+        + "numpy.testing.assert_allclose.html'. Passing a value for `decimal` will "
+        + "cause validation to revert to the old behaviour."
+    )
+    # decimal overrides new behaviour,
+    #   have to convert the params to emulate old behaviour
+    return {
+        "absolute_tolerance": 10 ** (-decimal),
```
FynnBe, line 438: I suppose to resemble previous behavior we need an additional factor of 1.5 (see https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_almost_equal.html)

```suggestion
        "absolute_tolerance": 1.5 * 10 ** (-decimal),
```

bioimageio/core/_resource_tests.py
```diff
@@ -361,6 +397,50 @@ def _test_expected_resource_type(
...
+        "relative_tolerance": 0,
+        "decimal": None,
```
FynnBe, line 440: I think we should not return "decimal" after handling the precision args
```suggestion
```

bioimageio/core/_resource_tests.py
```diff
@@ -110,7 +133,16 @@ def load_description_and_test(
         else:
             weight_formats = [weight_format]
         for w in weight_formats:
-            _test_model_inference(rd, w, devices, decimal)
+            # Note: new_precision_args is created like this to avoid type check errors
```
FynnBe, line 136: I think directly providing `absolute_tolerance` and  `relative_tolerance` to `_test_model_inference` would be a bit more straight forward...

bioimageio/core/_resource_tests.py
```diff
@@ -35,14 +36,21 @@ def test_model(
     source: Union[v0_5.ModelDescr, PermissiveFileSource],
     weight_format: Optional[WeightsFormat] = None,
     devices: Optional[List[str]] = None,
-    decimal: int = 4,
+    absolute_tolerance: float = 1e-4,
+    relative_tolerance: float = 1e-1,
```
FynnBe, line 40: 0.1 seems a bit too generous for a relative tolerance... especially considering we had 0 before and were mostly fine...
```suggestion
    relative_tolerance: float = 1e-4,
```

bioimageio/core/_resource_tests.py
```diff
@@ -361,6 +397,50 @@ def _test_expected_resource_type(
...
+    """
+    Arguments, both deprecated and current, for setting the precision during validation.
+    """
+
+    absolute_tolerance: float
+    relative_tolerance: float
+    decimal: Optional[int]
+
+
+def _handle_legacy_precision_args(
```
FynnBe, line 410: awesome idea to handle the legacy args separately with this helper!

FynnBe (changes requested):
Great start!

bioimageio/core/_resource_tests.py
melisande-c: Oops I think this was an automatically added import from my IDE before I switched to `TypedDict` for `PresicionArgs`, I will remove it

bioimageio/core/_resource_tests.py
melisande-c: The reason I did this is because a few different functions are called in a chain, so the precision arguments can be transformed in the first function called and passed onto the next function, which will also "transform" the arguments (but since they are already transformed they will be returned as they are). 

I suppose this transformation could just happen in the last function in the chain, but I wanted to make it explicit how these parameters are being handled for someone looking at a each individual function. This is also because I am not totally familiar with the code. If you think it is better, I can make the suggested change and only call `_handle_legacy_precision_args` in `load_description_and_test` or possibly `_test_model_inference`.

bioimageio/core/_resource_tests.py
melisande-c: Thoughts in this [comment](https://github.com/bioimage-io/core-bioimage-io-python/pull/418#discussion_r1738611678), I can make the change

bioimageio/core/_resource_tests.py
melisande-c: I can make the change, these were mostly placeholder defaults before getting some opinions.

I was going off the [numpy defaults](https://numpy.org/doc/stable/reference/generated/numpy.allclose.html) where the relative tolerance is `1e3` times less than the absolute tolerance. But I also chose it kinda arbitrarily.

FynnBe (approved):
Great! Thanks for the updates!

bioimageio/core/_resource_tests.py
```diff
@@ -35,13 +35,18 @@ def test_model(
     source: Union[v0_5.ModelDescr, PermissiveFileSource],
     weight_format: Optional[WeightsFormat] = None,
     devices: Optional[List[str]] = None,
-    decimal: int = 4,
+    absolute_tolerance: float = 1e-4,
```
FynnBe, line 38:
```suggestion
    absolute_tolerance: float = 1.5e-4,
```
to not be stricter than before...

bioimageio/core/_resource_tests.py
```diff
@@ -53,15 +58,20 @@ def test_description(
     format_version: Union[Literal["discover", "latest"], str] = "discover",
     weight_format: Optional[WeightsFormat] = None,
     devices: Optional[List[str]] = None,
-    decimal: int = 4,
+    absolute_tolerance: float = 1e-4,
```
FynnBe, line 61:
```suggestion
    absolute_tolerance: float = 1.5e-4,
```

bioimageio/core/_resource_tests.py
```diff
@@ -74,10 +84,13 @@ def load_description_and_test(
...
```
FynnBe, line 87:
```suggestion
    absolute_tolerance: float = 1.5e-4,
```

PR #452, opened 2025-02-20
Separates pipeline build logic per version

Tomaz-Vieira:
[This issue](https://github.com/bioimage-io/use-cases/issues/8) was due to `FixedZeroMeanUnitVariance` not being able to be created from a v4 preprocessing descriptor.

This PR splits the logic for v4 and v5 pipeline creation, making the types more visible and fixing the issue.

### DIFF
--- a/bioimageio/core/proc_ops.py
+++ b/bioimageio/core/proc_ops.py
@@ -3,7 +3,6 @@ from abc import ABC, abstractmethod
 from dataclasses import InitVar, dataclass, field
 from typing import (
     Collection,
-    List,
     Literal,
     Mapping,
     Optional,
@@ -19,6 +18,9 @@ from typing_extensions import Self, assert_never
 
 from bioimageio.spec.model import v0_4, v0_5
 from bioimageio.spec.model.v0_5 import TensorId
+from bioimageio.spec.model.v0_5 import (
+    _convert_proc,  # pyright: ignore [reportPrivateUsage]
+)
 
 from ._op_base import BlockedOperator, Operator
 from .axis import AxisId, PerAxis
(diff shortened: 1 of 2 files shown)


bioimageio/core/proc_setup.py
```diff
@@ -135,66 +141,60 @@ def get_requried_sample_measures(model: AnyModelDescr) -> RequiredSampleMeasures
...
+def _prepare_v4_preprocs(tensor_descrs: Sequence[v0_4.InputTensorDescr]) -> Tuple[List[Processing], Set[Measure]]:
+    procs: List[Processing] = []
+    for t_descr in tensor_descrs:
+        member_id = TensorId(str(t_descr.name))
+        procs.append(
+            EnsureDtype(input=member_id, output=member_id, dtype=t_descr.data_type)
+        )
+        for proc_d in t_descr.preprocessing:
+            procs.append(preproc_v4_to_processing(t_descr, proc_d))
+    measures = {m for proc in procs for m in proc.required_measures}
```
FynnBe, line 153: this we could maybe extract?
It's the same for all prepare funcs.

FynnBe (commented):
great stuff, thanks!

bioimageio/core/proc_setup.py
FynnBe: yes, just the measures

bioimageio/core/proc_ops.py
```diff
@@ -688,3 +692,113 @@ def get_proc_class(proc_spec: ProcDescr):
...
+                v5_proc_spec, v0_5.FixedZeroMeanUnitVarianceDescr
+            )  # FIXME
+            return FixedZeroMeanUnitVariance.from_proc_descr(v5_proc_spec, member_id)
+        else:
+            return ZeroMeanUnitVariance.from_proc_descr(proc_spec, member_id)
+    else:
+        assert_never(proc_spec)
+
+
+def preproc_v5_to_processing(
```
FynnBe, line 757: Might be a bit pedantic..., but couldn't we just merge `preproc_v5... ` and `postproc_v5...` and annotate with the union?
Those two have a lot of code duplication and pre and postproc descr are just different unions of the same set of descriptions.

Same for v4.

Tomaz-Vieira: Fair point. I guess I was thinking that there would be both `Preprocessing` and `Postprocessing` classes, instead of just `Processing`. I unified them now

Tomaz-Vieira:
Yup, looks about right. I'll close this PR, then

Issue #477, opened 2026-01-08
consider cpu fallback for apple silicon devices


k-dominik:
pytorch does not support all operations on the accelerated backend on macos "mps" (metal performance shader). The safe alternative is running things on the cpu backend - but that's slow. pytorch provides an environment variable that should enable always using the "mps" backend and automatically falling back to cpu for unsupported operations. In ilastik we have the following snippet, which starts a suprocess running inference with a modified env:

```Python
    @property
    def _proc_env(self):
        """Modified env for tiktorch server subprocess"""
        env = os.environ.copy()
        if platform.system().lower() == "darwin" and platform.machine().lower() == "arm64":
            # pytorch versions up to the 2.6 don't support all operations (esp 3d) on MPS
            # this env variable allows falling back to CPU for those networks instead of failing
            # see for current status https://github.com/pytorch/pytorch/issues/141287
            if "PYTORCH_ENABLE_MPS_FALLBACK" not in env:
                env["PYTORCH_ENABLE_MPS_FALLBACK"] = "1"
                logging.info(f"Added 'PYTORCH_ENABLE_MPS_FALLBACK=1' to tiktorch env.")

        return env
```

https://github.com/ilastik/ilastik/blob/59370167576d7c8cf8c789f75effc0f1935c2e39/ilastik/workflows/neuralNetwork/_localLauncher.py#L63-L75

xref: https://qqaatw.dev/pytorch-mps-ops-coverage/

FynnBe:
done

PR #481, opened 2026-04-01
enable mps pytorch backend if available and some housekeeping

FynnBe:


FynnBe:
Closes #477

k-dominik (approved):
lgtm. Note that it seems that in pytorch 3.11 (not available yet) A LOT of the stuff seems now implemented - having said that I don't think it hurts to set this env variable anyway as it helps  the old versions to at least produce an output rather than failing!

No clue about the early stopping you want to do there ;)

src/bioimageio/core/_settings.py
```diff
@@ -13,6 +16,27 @@ class Settings(SpecSettings):
...
+        # pytorch versions up to the 2.6 don't support all operations (esp 3d) on MPS
+        # this env variable allows falling back to CPU for those networks instead of failing
+        # see for current status https://github.com/pytorch/pytorch/issues/141287
+        if (
+            value is None
+            and platform.system().lower() == "darwin"
+            and platform.machine().lower() == "arm64"
+        ):
+            logger.info("Set environment variable 'PYTORCH_ENABLE_MPS_FALLBACK=1'.")
+            os.environ["PYTORCH_ENABLE_MPS_FALLBACK"] = "1"
```
k-dominik, line 35: ah hang on - in the version I had, the user could still override this env variable, which I think is the proper thing to do.

k-dominik (changes requested):

k-dominik (approved):

src/bioimageio/core/_settings.py
FynnBe: yes, set env vars are respected by pytdantic.BaseSettings. It's only set in this after field validator if not in the digested env vars

PR #489, opened 2026-04-22
feat: CustomPostprocessing op — user-supplied callable via source file

oeway:
## Summary

Implements the runtime side of the new `CustomPostprocessingDescr` spec field (spec PR bioimage-io/spec-bioimage-io#752).

### Design

Model contributors package a Python source file alongside their weights and reference it in `rdf.yaml`:

```yaml
postprocessing:
  - id: custom
    callable: my_postprocess
    source: my_postprocess.py
    sha256: <hash>
    kwargs:
      threshold: 0.5
```

At runtime, `bioimageio.core` loads and SHA-256-verifies the file, then runs:
```python
op = callable(**kwargs)   # once at construction
result = op(*tensors)     # once per sample
```

Both callable-class and factory-function styles are supported.

### Changes

**`proc_ops.py`**
- `CustomPostprocessing(SamplewiseOperator)` — loads source, verifies hash, constructs and caches the user callable
- Error branch in `get_proc` to guide callers toward `proc_setup` (needs all output ids)
- Added to `Processing` union

**`proc_setup.py`**
- `_get_described_procs` computes `all_output_ids` upfront and dispatches `CustomPostprocessingDescr` directly to `CustomPostprocessing.from_proc_descr`, passing all output member ids

**`custom_ops/`**
- New directory for shared built-in ops contributed by the community
- `cellpose_flow_dynamics.py` — first op: decodes Cellpose/Cellpose-SAM flow fields into instance labels
- `README.md` — contributor guide

### Promotion path

Once a contributed op is accepted into `custom_ops/`, it gets its own named `id` in the spec and models no longer need `source`/`sha256`/`callable`.

cc @FynnBe


oeway:
@FynnBe CI is green on this PR — ready for review and merge.

Summary of changes:
- `CustomPostprocessing(SamplewiseOperator)` in `proc_ops.py`: loads source from bytes via tempfile + importlib, instantiates callable once with kwargs, calls per sample
- `proc_setup.py` dispatches `CustomPostprocessingDescr` (guarded with `getattr` for older spec installs)
- `custom_ops/cellpose_flow_dynamics.py`: built-in Cellpose flow-dynamics op
- `custom_ops/README.md`: contributor guide

Note: `coverage-comment` step failed (permissions issue posting to PR), all actual tests and pyright passed.

Depends on companion PR bioimage-io/spec-bioimage-io#752.

FynnBe:
@oeway you write that once an op is added to core.custom_ops it gets an id, but cellpose_flow_dynamics is not getting an id in this PR.
An op can't be built-in and custom at the same time I'd say, so I'll go ahead and add cellpose postprocessing as a known op (not a custom one)

oeway:
@FynnBe cellpose_flow_dynamics is an example, feel free to make changes and move a head with merging! Thanks!

FynnBe:
I'm adding explicit padding at the sample edge to be able to match cellpose behavior.

When using blockwise prediction cellpose is averaging across the tile overlap, while we do not and expect the cropped halo to be sufficiently large to avoid blocking artifacts, which holds mostly true for CNNs, but not necessarily for transformers.
I plan to add block-averaging in a subsequent PR/release and live with potential block artifacts for now as they can be mediated to some extend by using larger halos.

FynnBe:
went back to pyright 1.1.408 as torch imports were flagged as private, see https://github.com/microsoft/pyright/issues/11405

Issue #491, opened 2026-05-13
GitHub releases not updated despite new tags / feedstock releases


anwai98:
Hi team,

While debugging some `micro-sam` related stuff, I realized that `spec` and `core` have been updated super recently, but the GitHub releases aren't updated (although the release tags and feedstock seems to be updated).

Is there anyway the release drafter can be fixed? (this is my best guess that something isn't working there, but didn't dig into the workflows)

FynnBe:
Hi @anwai98, 
yes, I suspect the same. I will take another look at fixing the github releases. If you have a suspicion about what is going wrong, please share!
[The workflow](https://github.com/bioimage-io/core-bioimage-io-python/actions/runs/25758247014/job/75652328950) looks like it runs successfully, but the release page begs to differ...

anwai98:
Hi @FynnBe,

Gotcha. I'll check it out in a few days (I spotted the release drafter workflow).

FynnBe:
somehow the workflow is only updating the draft. Apparently the draft then can be released manually, but I would prefer this to be done automatically...

anwai98:
Hi @FynnBe,

I see the release drafter already open. Seems like it works ;)

I'll go ahead and close this issue!

anwai98:
Hi @FynnBe,

I just noticed that the release drafter is gone now, and the new bump didn't get released either. Do you know what might have happened there? (probably similar things happened in `spec` too)

FynnBe:
It looked to me like only [this line](https://github.com/bioimage-io/core-bioimage-io-python/blob/e783f53e88dc293348c1be7bc66fa5d2f41c5c88/.github/workflows/build.yaml#L351) was missing... 
Your changes to the dedicated release drafter workflow unfortunately broke the release pipeline that itself pushes a new version tag.
I'm leaving this issue open until I verified that the `publish: true` addition in the CI workflow fixes the release note publishing.

anwai98:
Hi @FynnBe,

Re: broken release pipeline: I see.

> I'm leaving this issue open until I verified that the publish: true addition in the CI workflow fixes the release note publishing.

Okay, sounds good. I have a few more hunches how we could get the release drafter up and running, but I'll let you check the `publish: true` workflow first!

FynnBe:
it worked!

anwai98:
Awesome. That sounds great. Thanks @FynnBe! 🥳

PR #492, opened 2026-05-14
Fix release drafter by adding permission and auto publish on version bump

anwai98:
Hi @FynnBe,

I think the drafter should be working now. It was previously missing `contents: write`, and `pull-requests: write` sort of permissions, so my best guess is this is one of the reasons why the action silently couldn't create or update release drafts.

Also, I noticed that `release-drafter.yml` was missing `name-template`, `tag-template`, and `version-resolver`, so the action had no way to name or version **releases.**

Now regarding the manual part: I tagged a version change detector, i.e. bumping `__version__` in `src/bioimageio/core/__init__.py` and merging to `main` now should in theory automatically tag the commit and publish the release draft. Maybe good to keep in mind that non-bumping merges would continue to accumulate the draft.

PS. Most of the ideas here are hacked up from our workflow in `micro-sam` for this ;)

I think this should be working, however can't test extensively at my end, unless the next rc is on-the-way!

Let me know how it looks! (and please feel free to take over and change things as you like)

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,3 +1,25 @@
+name-template: 'v$RESOLVED_VERSION'
+tag-template: 'v$RESOLVED_VERSION'
+version-resolver:
+  major:
+    labels:
+      - 'breaking'
+  minor:
+    labels:
+      - 'enhancement'
+  patch:
+    labels:
+      - 'bug'
+      - 'performance'
+      - 'testing'
+      - 'ci'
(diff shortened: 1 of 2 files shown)


anwai98:
And in theory, closes https://github.com/bioimage-io/core-bioimage-io-python/issues/491

FynnBe:
Thank you @anwai98 ! Let's give it a try...

PR #501, opened 2026-08-17
Add micro-sam (μSAM) watershed postprocessing

nilsmechtel:
## Motivation

micro-sam models with the additional instance segmentation (AIS) decoder predict three dense maps (foreground probability, center distance, inverted boundary distance). Exported as bioimageio packages, such models run a plain prompt-free prediction pass, but turning the maps into instance labels requires micro-sam's seeded watershed, which so far only exists inside the `micro_sam` package. This PR makes that final step available to any tool building on `bioimageio.core`, following the precedent of the stardist and cellpose postprocessing already in this repo.

## What is added

- `MicroSamWatershed`, a `SamplewiseOperator` that replaces the three stacked decoder maps on its sample member with instance labels. Seeds are connected components where both smoothed distance maps are below their thresholds inside the foreground mask; the seeded watershed runs on the smoothed boundary distances restricted to that mask. Faithful port of `micro_sam.instance_segmentation.InstanceSegmentationWithDecoder` / `torch_em.util.segmentation.watershed_from_center_and_boundary_distances`, with the upstream parameter defaults (thresholds 0.5, foreground smoothing 1.0, distance smoothing 1.6, `min_size` 0).
- `from_proc_descr(MicroSamWatershedDescr, member_id)` plus registration in `proc_ops` — import, `Processing` union, and a `get_proc` dispatch branch — mirroring `CellposeFlowDynamics`.
- Optional extra `bioimageio.core[microsam]` providing `scikit-image` (only needed for the watershed itself; gaussian smoothing and connected components use the existing scipy dependency), mirroring the `stardist` extra. scikit-image is imported lazily so the module stays importable without it.
- Unit tests with synthetic two-cell maps: instance separation across a boundary ridge, `min_size` filtering, channel-count validation, and `from_proc_descr` kwargs mapping.

## Depends on bioimage-io/spec-bioimage-io#768

`from_proc_descr` binds to `MicroSamWatershedDescr`, which that PR adds to `bioimageio.spec.model.v0_5`. Until it is released, the `bioimageio.spec ==0.5.12.0` pin here does not contain the descriptor and CI cannot pass; the pin needs bumping to whichever release carries it. Verified locally against a co-developed spec checkout (`extraPaths = ["../spec-bioimage-io/src"]`).

## Test plan

- `pytest tests/test_ops_microsam.py` — 4 passed, against the spec branch of #768.
- `get_proc` dispatch checked end to end: an `OutputTensorDescr` carrying a `MicroSamWatershedDescr` postprocessing entry resolves to a `MicroSamWatershed` with the declared kwargs and the tensor's member id.
- Behavior cross-checked against the micro_sam / torch_em source and the parameter defaults used by the micro-sam BioEngine deployment.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,6 +52,7 @@ partners = [
     # "stardist",         # for model testing and stardist postprocessing  # TODO: add updated stardist to partners env
 ]
 stardist = ["stardist"] # for stardist postprocessing
+microsam = ["scikit-image"] # for micro-sam (μSAM) watershed postprocessing
 gradio-server = ['gradio[mcp];python_version>="3.10"']
 gradio-client = ['gradio_client;python_version>="3.10"']
 dev = [
--- a/src/bioimageio/core/__init__.py
+++ b/src/bioimageio/core/__init__.py
@@ -55,6 +55,7 @@ from . import stat_measures as stat_measures
 from . import tensor as tensor
 from . import weight_converters as weight_converters
 from ._prediction_pipeline import IntermediatePrediction as IntermediatePrediction
+from ._ops_microsam import microsam_watershed as microsam_watershed
 from ._prediction_pipeline import PredictionPipeline as PredictionPipeline
 from ._prediction_pipeline import RemotePredictionPipeline as RemotePredictionPipeline
 from ._prediction_pipeline import (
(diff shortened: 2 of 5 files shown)


FynnBe:
just some guidance: Any added op should inherite from `SamplewiseOperator` or `BlockwiseOperator` and should have a `from_proc_descr` method that is called with an appropriate bioimageio.spec.model.v0_5 object.

